PR #1, opened 2022-06-30
Create common sdk Fastfile

tonidero:
https://revenuecats.atlassian.net/browse/CSDK-27

Added a `Fastfile` containing all the shared automation logic between our SDKs. This PR is just moving the `CommonFastfile` in iOS that was added here: https://github.com/RevenueCat/purchases-ios/pull/1719.

Once this is merged, I will remove the one in the iOS repo and import this one directly.

### DIFF
--- a/sdks/fastlane/Fastfile
+++ b/sdks/fastlane/Fastfile
@@ -1,0 +1,274 @@
+before_all do
+  update_fastlane
+end
+
+desc "Bump version, edit changelog, and create pull request"
+lane :bump_version_update_changelog_create_pr do |options|
+  ensure_git_branch(branch: options[:branch] || 'main')
+  ensure_git_status_clean
+
+  # Ensure GitHub API token is set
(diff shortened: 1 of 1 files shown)


joshdholtz:
@tonidero  Just wanted to leave a thought here about this! There are two approaches we could use:

1. Share a `Fastfile`
2. Create a [plugin](https://docs.fastlane.tools/plugins/create-plugin/#create-your-own-fastlane-plugin) with a bunch of shared actions (where the actions are the essentially the shared lanes)

My original thought was to make a plugin 😇 One benefit to the plugin is that all the actions that are defined in it can be documented and all the options are named, validated, and documented. I think this would make maintaining and using these actions across all the SDKs the easiest and cleanest in the end.

On the other hand, a shared `Fastfile` does allow for passing in options but its a un-typed and un-validated hash/dictionary and could be more error prone.

cc: @aboedo

PR #2, opened 2022-07-04
Create plugin

tonidero:
This PR creates the plugin with the initial actions that will help to handle version updates, changelog updates and PR creation for releases. The actions added in this PR are:
- `bump_version_update_changelog_create_pr`. This action will be used by SDKs to create new releases
- `replace_version_number`. This action is part of the process `bump_version_update_changelog_create_pr`. I added this action and made it another action as a way to test more easily + in case devs want to do things more manually. Generally, SDKs will not end up using this action directly though.

There are a few remaining actions that I will move to the plugin in future PRs.

This plugin was tested on the `purchases-android` sdk, and was able to create a test release PR successfully: https://github.com/RevenueCat/purchases-android/pull/556


.circleci/config.yml
```diff
@@ -0,0 +1,43 @@
+# Ruby CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-ruby/ for more details
+#
+version: 2.1
+jobs:
+  build:
+    docker:
+      # specify the version you desire here
+       - image: circleci/ruby:3.1.2
```
tonidero, line 10: I'm using this locally, so I updated to this, though we could use the min supported ruby version here (2.5)

.gitignore
```diff
@@ -0,0 +1,12 @@
+*.gem
+Gemfile.lock
```
tonidero, line 2: I'm still not sure if we should gitignore the `Gemfile.lock`... But this was part of the default output from `fastlane` so I decided to leave it for now.

Rakefile
```diff
@@ -0,0 +1,11 @@
+require 'bundler/gem_tasks'
+
+require 'rspec/core/rake_task'
+RSpec::Core::RakeTask.new
+
+# Not using RuboCop::RakeTask.new(:rubocop) since it's showing different errors than just running rubocop
```
tonidero, line 6: The default was to use this `RuboCop::RakeTask.new(:rubocop)` to create the rake task, but that was showing a different output than just running `rubocop` or `bundle exec rubocop`. I checked and the version was the same in both cases, so I didn't know why the output was different... When I run it with this task, I got the following errors:
```
lib/fastlane/plugin/revenuecat/actions/bump_version_update_changelog_create_pr_action.rb:56:11: C: Require/MissingRequireStatement: FastlaneCore::ConfigItem not found, you're probably missing a require statement or there is a cycle in your dependencies.
          FastlaneCore::ConfigItem.new(key: :current_version,
          ^^^^^^^^^^^^^^^^^^^^^^^^
lib/fastlane/plugin/revenuecat/actions/replace_version_number_action.rb:38:11: C: Require/MissingRequireStatement: FastlaneCore::ConfigItem not found, you're probably missing a require statement or there is a cycle in your dependencies.
          FastlaneCore::ConfigItem.new(key: :current_version,
          ^^^^^^^^^^^^^^^^^^^^^^^^
lib/fastlane/plugin/revenuecat/helper/revenuecat_helper.rb:15:9: C: Require/MissingRequireStatement: Action not found, you're probably missing a require statement or there is a cycle in your dependencies.
        Action.sh("sed", '-i', backup_extension, sed_regex, path)
        ^^^^^^
```
But when running rubocop directly, it didn't show any errors. Also, tests run fine, so I decided to just run `rubocop` directly instead of using this precreated rake task.

fastlane-plugin-revenuecat.gemspec
```diff
@@ -0,0 +1,36 @@
...
+
+  spec.required_ruby_version = '>= 2.5'
+
+  # Don't add a dependency to fastlane or fastlane_re
+  # since this would cause a circular dependency
+
+  # spec.add_dependency 'your-dependency', '~> 1.0.0'
+
+  spec.add_development_dependency('bundler')
+  spec.add_development_dependency('fastlane', '2.191.0')
```
tonidero, line 26: I had to downgrade fastlane to this version... If I run it with `2.192.0` or greater, I get the following error:
```
An error occurred while loading spec_helper. - Did you mean?
                    rspec ./spec/spec_helper.rb

Failure/Error: require 'fastlane' # to import the Action super class

Gem::ConflictError:
  Unable to activate fastlane-2.207.0, because optparse-0.2.0 conflicts with optparse (~> 0.1.1)
# ./spec/spec_helper.rb:12:in `<top (required)>'
# ------------------
# --- Caused by: ---
# LoadError:
#   cannot load such file -- fastlane
#   ./spec/spec_helper.rb:12:in `<top (required)>'
No examples found.
No examples found.
```

fastlane-plugin-revenuecat.gemspec
```diff
@@ -0,0 +1,36 @@
...
+  spec.add_development_dependency('pry')
+  spec.add_development_dependency('rake')
+  spec.add_development_dependency('rspec')
+  spec.add_development_dependency('rspec_junit_formatter')
+  spec.add_development_dependency('rubocop', '1.31.1')
```
tonidero, line 31: I updated this version to the latest rubocop version to solve some issues I was getting

lib/fastlane/plugin/revenuecat/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -0,0 +1,110 @@
+require 'fastlane/action'
+require 'fastlane_core/configuration/config_item'
+require_relative '../helper/revenuecat_helper'
+
+module Fastlane
+  module Actions
+    class BumpVersionUpdateChangelogCreatePRAction < Action
+      def self.run(params)
+        # TODO
```
tonidero, line 9: Still in progress with this action, will probably be the last one since it's the more complex action.

lib/fastlane/plugin/revenuecat/actions/replace_version_number_action.rb
```diff
@@ -0,0 +1,65 @@
...
+
+module Fastlane
+  module Actions
+    class ReplaceVersionNumberAction < Action
+      def self.run(params)
+        previous_version_number = params[:current_version]
+        new_version_number = params[:new_version]
+        files_to_update = params[:files_to_update]
+        files_to_update_without_prerelease_modifiers = params[:files_to_update_without_prerelease_modifiers]
+        UI.user_error!("missing current version param") unless previous_version_number
```
tonidero, line 13: I'm not sure if this is needed... I would think fastlane would add some input validation using the `available_options` method but I couldn't get that to work on the tests, so I added some basic validation here for now... It would be great to get some intro into how the `available_options` work when you have time @joshdholtz

lib/fastlane/plugin/revenuecat/helper/revenuecat_helper.rb
```diff
@@ -0,0 +1,156 @@
...
+        last_version = Actions.sh("git describe --tags --abbrev=0").strip
+        old_version ||= last_version
+        UI.important("Auto-generating changelog since #{old_version}")
+
+        org = "RevenueCat"
+
+        path = "/repos/#{org}/#{repo_name}/compare/#{old_version}...HEAD"
+
+        # Get all commits from previous version (tag) to HEAD
+        resp = github_api(path: path, api_token: github_token)
```
tonidero: Hey @joshdholtz! I was running into a new issue, in case you have any suggestions about this :sweat_smile:. I'm trying to move the logic to auto generate a changelog based on the commits since the last release to the new plugin. This code currently uses the `github_api` action, as seen here. 

I moved this code out of the action itself and into the helper, to keep the action cleaner. When I try to call `github_api` directly like this (code not working currently), it can't find the action. It does find it if I do `Actions::GithubApiAction.run`, but doing it that way, I don't get the input validation, nor the default params on that action. I tried using `other_action.github_api`, but it seems it can't find `other_action` in the helper (tried a few things without luck).

Should I move this logic back to the `action` so I can use `other_action`, or do you know a better way to call other actions from the helper?

fastlane-plugin-revenuecat.gemspec
tonidero: I was able to fix this by downgrading my ruby for this project. Now I can use the latest fastlane version for this.

lib/fastlane/plugin/revenuecat/actions/replace_version_number_action.rb
tonidero: I tested the plugin from a project and it indeed does the input validation, so I'm removing this validation here. I'm a bit concerned for not being able to test that same input from the tests though...

Rakefile
joshdholtz: Weird that it didn't show the same errors! Not sure why that would happen 🤷‍♂️ This works so that is what matters 😊

fastlane-plugin-revenuecat.gemspec
```diff
@@ -0,0 +1,36 @@
+lib = File.expand_path("lib", __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'fastlane/plugin/revenuecat/version'
+
+Gem::Specification.new do |spec|
+  spec.name          = 'fastlane-plugin-revenuecat'
```
joshdholtz, line 6: Thoughts on a potential rename to something like `fastlane-plugin-revenuecat-internal` along with a repo rename? 😅 

Was just thinking that we _may_ want to keep the `fastlane-plugin-revenuecat` namespace and repo available for if/when we ever create a _fastlane_ plugin for our customers to use 🤷‍♂️

lib/fastlane/plugin/revenuecat/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -0,0 +1,109 @@
...
+          FastlaneCore::ConfigItem.new(key: :changelog_latest_path,
+                                       description: "Path to CHANGELOG.latest.md file",
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :changelog_path,
+                                       description: "Path to CHANGELOG.md file",
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :files_to_update,
+                                       env_name: "FILES_TO_UPDATE_VERSION",
```
joshdholtz, line 64: Not sure if this is documented anywhere in _fastlane_ (but I should probably add it), but it is best practice for `env_name` to be prefixed with the name of the action so that they don't get overridden or accidentally used by other actions.

I think these naming are probably fine and won't overlap but maybe we could prefix all of these `env_name`s with something like `RC_` (or `RC_INTERNAL`) if we go with a plugin rename?

lib/fastlane/plugin/revenuecat/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -0,0 +1,109 @@
...
+                                       env_name: "GITHUB_TOKEN",
+                                       description: "Github token to use to prepopulate the changelog",
+                                       optional: true,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :github_rate_limit,
+                                       env_name: "GITHUB_RATE_LIMIT_SLEEP",
+                                       description: "Sets a rate limiter for github requests when creating the changelog",
+                                       optional: true,
+                                       default_value: 0,
+                                       type: Integer),
```
joshdholtz, line 89: I need to redo how the changelog generator works so that the isn't needed 🙈

joshdholtz (changes requested):
This looks really good! I'm going to love having this 😊 

Left a few suggestions but none I _think_ are really major but I'd love to hear your thoughts!

fastlane-plugin-revenuecat.gemspec
tonidero: That makes a lot of sense 👍 Will do the rename in a bit.

And what do you think about publishing this plugin in RubyGems? I honestly thought it might be a better idea to have the SDKs just point to github, since this is only going to be internal, but if you think we would benefit in any way from publishing I can work on that

lib/fastlane/plugin/revenuecat/actions/bump_version_update_changelog_create_pr_action.rb
tonidero: When I was adding tests for that part of the process, I noticed why we have to do a request for each commit. To get the PR number. I think all the other information would be available within the response of the list of commits. I'm not sure if there is an easier way to get the PR number from each commit, but it would be great to avoid this problem :P

tonidero:
This seems to be working now! This is a generated release PR created using the plugin: https://github.com/RevenueCat/purchases-android/pull/556. Will clean this up, write a few more tests and mark it ready for review.

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -0,0 +1,122 @@
...
+                                       optional: true,
+                                       default_value: 0,
+                                       type: Integer),
+          FastlaneCore::ConfigItem.new(key: :branch,
+                                       description: "Allows to execute the action from the given branch",
+                                       optional: true,
+                                       default_value: "main",
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :editor,
+                                       env_name: "RC_INTERNAL_FASTLANE_EDITOR",
```
tonidero, line 109: Note that in `purchases-ios` we currently use either `FASTLANE_EDITOR` or `EDITOR` as environment variables to override this. So devs will need to update this in case they want to use their own editor.

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -0,0 +1,122 @@
...
+                                       optional: true,
+                                       default_value: [],
+                                       type: Array),
+          FastlaneCore::ConfigItem.new(key: :repo_name,
+                                       env_name: "RC_INTERNAL_REPO_NAME",
+                                       description: "Name of the repo of the SDK",
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :github_pr_token,
+                                       env_name: "GITHUB_PULL_REQUEST_API_TOKEN",
```
tonidero, line 88: I thought about renaming this env within the `RC_INTERNAL_` namespace as well... but I felt I could reuse the one that is used by the `create_pull_request` fastlane action... Lmk if you think I should rename this. It would also mean ios devs need to update their ENVs to use the new name.

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -0,0 +1,155 @@
...
+      def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
+        old_version = Actions.sh("git describe --tags --abbrev=0").strip
+        UI.important("Auto-generating changelog since #{old_version}")
+
+        org = "RevenueCat"
+
+        path = "/repos/#{org}/#{repo_name}/compare/#{old_version}...HEAD"
+
+        # Get all commits from previous version (tag) to HEAD
+        resp = Actions::GithubApiAction.run(server_url: 'https://api.github.com',
```
tonidero, line 35: I've had to use a lot of actions with this form. I'm not a fan since we don't get the param validation or the default values for these params... But wasn't able to call an action from the helper in any other way. Lmk if you have any recommendations.

lib/fastlane/plugin/revenuecat_internal/version.rb
```diff
@@ -0,0 +1,5 @@
+module Fastlane
+  module RevenuecatInternal
+    VERSION = "0.1.0"
```
tonidero, line 3: Since this is internal, I think we can keep it like this for now. My idea is to import this plugin in the SDKs using github directly instead of RubyGems, but I'm open for discussion.

tonidero:
Work supporting this plugin on `purchases-android` is being done here: https://github.com/RevenueCat/purchases-android/pull/557

fastlane-plugin-revenuecat_internal.gemspec
```diff
@@ -0,0 +1,36 @@
...
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'fastlane/plugin/revenuecat_internal/version'
+
+Gem::Specification.new do |spec|
+  spec.name          = 'fastlane-plugin-revenuecat_internal'
+  spec.version       = Fastlane::RevenuecatInternal::VERSION
+  spec.author        = 'Toni Rico'
+  spec.email         = 'toni.rico.diez@revenuecat.com'
+
+  spec.summary       = 'A plugin including commonly used automation logic for RevenueCat SDKs.'
```
NachoSoto, line 11: 👏🏻

fastlane/Fastfile
```diff
@@ -0,0 +1,24 @@
+lane :test_bump_version_update_changelog_create_pr do |options|
```
NachoSoto, line 1: Is this just to test them locally in the repo?

spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      allow(Fastlane::Helper::RevenuecatInternalHelper).to receive(:auto_generate_changelog).with(anything).and_return(auto_generated_changelog)
+      allow(Fastlane::Helper::RevenuecatInternalHelper).to receive(:edit_changelog).with(anything)
+      allow(Fastlane::Helper::RevenuecatInternalHelper).to receive(:create_new_release_branch).with(anything)
+      allow(Fastlane::Helper::RevenuecatInternalHelper).to receive(:replace_version_number).with(anything)
+      allow(Fastlane::Helper::RevenuecatInternalHelper).to receive(:attach_changelog_to_master).with(anything)
+      allow(Fastlane::Helper::RevenuecatInternalHelper).to receive(:commmit_changes_and_push_current_branch).with(anything)
+      allow(Fastlane::Helper::RevenuecatInternalHelper).to receive(:create_release_pr).with(anything)
+    end
+
+    it 'calls all the appropriate methods with appropriate parameters' do
```
NachoSoto, line 28: Whoa even with tests? 🤯

NachoSoto (commented):
Awesome :clap

fastlane/Fastfile
tonidero: These are mostly as an example of how to call the different actions. If we currently call them, they will fail, since the parameters passed do not match the project setup. I think I can rename these lanes to `sample_bump_version_update_changelog_create_pr_call` or something like that

spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
tonidero: IMO, this is the main reason for moving this into a plugin ❤️

tonidero:
Bump @RevenueCat/sdk

LICENSE
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
taquitos, line 21: What we use is:

//
//  Copyright RevenueCat Inc. All Rights Reserved.
//
//  Licensed under the MIT License (the "License");
//  you may not use this file except in compliance with the License.
//  You may obtain a copy of the License at
//
//      https://opensource.org/licenses/MIT
//
//  <filename>
//
//  Created by <Full name if you want> on <mm/dd/yy>
//

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
taquitos: I'm cool with keeping this 😄

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
taquitos: CC @joshdholtz

lib/fastlane/plugin/revenuecat_internal/version.rb
taquitos: I think that makes sense. If we made it public, it feels like RubyGems would make more sense then.

spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
taquitos: Yeah this is dope

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -0,0 +1,306 @@
...
+      expect(changelog_contents).to eq("## 1.12.0\nchangelog latest contents\n## 1.11.0\nchangelog contents")
+    end
+  end
+
+  describe '.create_new_release_branch' do
+    it 'creates new release branch with version number' do
+      expect(Fastlane::Actions).to receive(:sh).with("git checkout -b 'release/1.12.0'")
+      Fastlane::Helper::RevenuecatInternalHelper.create_new_release_branch('1.12.0')
+    end
+  end
```
taquitos, line 231: I think you could do another test that checks if the branch already exists on remote. If it does, we should at least halt deployment.

taquitos (changes requested):
This looks really great! Just a few comments

LICENSE
tonidero: Hmm that seems like the header we add to files in iOS. This is a `LICENSE` file which I believe is supposed to be more generic, explaining the license itself... I checked and we have the same `LICENSE` on the iOS and Android SDKs... That said, I'm not fully familiar with this, so happy to talk about it if you think we should change it!

spec/helper/revenuecat_internal_helper_spec.rb
tonidero: Right now I was just copying the existing logic from iOS (which has the most advanced automation IMO) where it wouldn't fail in that case but I can add that logic since it shouldn't be difficult. Will do so in another PR though if that's ok so it can be reviewed better. I will do that at the beginning of the`BumpVersionUpdateChangelogCreatePrAction` since it's where we create the release branch.

tonidero: Added this validation here: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/5

LICENSE
taquitos: Sorry, I didn't see that, I thought it was a header in a file! You're right.
We still need to change it and attribute it back to RC 😄
`Copyright (c) 2022 Toni Rico <antonio.rico.diez@revenuecat.com>` -> `Copyright RevenueCat Inc. All Rights Reserved.`

spec/helper/revenuecat_internal_helper_spec.rb
taquitos: Love it, thank you!

taquitos (approved):
image

tonidero:
@joshdholtz Can you take another look when you have time? Lmk if you don't have time and I can bypass merge requirements.

tonidero:
Going to merge this one in for now and I can address any other comments in future PRs. And sorry for the huge PR!

PR #3, opened 2022-07-08
Add create_next_snapshot_version action to plugin

tonidero:
This PR adds a new action to the fastlane plugin, `create_next_snapshot_version`. This action will bump the version to the next minor with a `-SNAPSHOT` version and create a PR with the changes.

### TODO
- [x] Hold until #2 is merged

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
@@ -18,14 +18,7 @@ module Fastlane
         changelog_path = params[:changelog_path]
         editor = params[:editor]
 
-        # Ensure GitHub API token is set
-        if github_pr_token.nil? || github_pr_token.empty?
-          UI.error("A github_pr_token parameter or an environment variable GITHUB_PULL_REQUEST_API_TOKEN is required to create a pull request")
-          UI.error("Please make a fastlane/.env file from the fastlane/.env.SAMPLE template")
-          UI.user_error!("Could not find value for GITHUB_PULL_REQUEST_API_TOKEN")
-        end
-
-        Helper::RevenuecatInternalHelper.validate_local_config_status_for_bump(branch)
+        Helper::RevenuecatInternalHelper.validate_local_config_status_for_bump(branch, github_pr_token)
 
         UI.important("Current version is #{version_number}")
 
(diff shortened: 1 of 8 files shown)


tonidero:
bump @RevenueCat/sdk

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -36,14 +29,14 @@ def self.run(params)
...
-        Helper::RevenuecatInternalHelper.create_new_release_branch(new_version_number)
+        Helper::RevenuecatInternalHelper.create_and_checkout_new_branch("release/#{new_version_number}")
         Helper::RevenuecatInternalHelper.replace_version_number(version_number,
                                                                 new_version_number,
                                                                 files_to_update,
                                                                 files_to_update_without_prerelease_modifiers)
         Helper::RevenuecatInternalHelper.attach_changelog_to_master(new_version_number, changelog_latest_path, changelog_path)
         Helper::RevenuecatInternalHelper.commmit_changes_and_push_current_branch("Version bump for #{new_version_number}")
-        Helper::RevenuecatInternalHelper.create_release_pr(new_version_number, changelog, repo_name, github_pr_token)
+        Helper::RevenuecatInternalHelper.create_pr_to_main("Release/#{new_version_number}", changelog, repo_name, github_pr_token)
```
aboedo, line 39: we don't have any rules for ruby code yet, but I wonder if we should start using named params. 
It took me a second to understand that the reason that on line 32 we do `release/` (lowercase) and then on 39 `Release/` is that one is the branch name and the other one the PR title

lib/fastlane/plugin/revenuecat_internal/actions/create_next_snapshot_version_action.rb
```diff
@@ -0,0 +1,81 @@
...
+                                       env_name: "RC_INTERNAL_FILES_TO_UPDATE_VERSION",
+                                       description: "Files that contain the version number and need to have it updated",
+                                       optional: false,
+                                       type: Array),
+          FastlaneCore::ConfigItem.new(key: :files_to_update_without_prerelease_modifiers,
+                                       env_name: "RC_INTERNAL_FILES_TO_UPDATE_VERSION_WITHOUT_PRERELEASE_MODIFIERS",
+                                       description: "Files that contain the version number without release modifiers and need to have it updated",
+                                       optional: true,
+                                       default_value: [],
+                                       type: Array),
```
aboedo, line 67: for this specific command, this will effectively mean "files for which you don't want to add the `-SNAPSHOT` suffix, right? 
Maybe we could name this param accordingly?

aboedo (approved):
looks awesome!! 🚢

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
tonidero: Hmm it's a good point... I'm a bit reticent doing it for methods with few parameters, instead we can store the parameter in a properly named variable (as I will do in this case) but I totally agree for methods with a lot of parameters. In this case with 4 parameters, I'm a bit on the fence... I would say to do it with 5+ parameters but I would be happy to discuss a different number. Lmk what you think!

lib/fastlane/plugin/revenuecat_internal/actions/create_next_snapshot_version_action.rb
tonidero: I wanted it to share the name of the similar parameter of the `BumpVersionUpdateChangelogCreatePrAction`. Even though both actions don't do the same thing, the behavior will be similar, so I think it's better to keep it consistent... I could rename it to something like `files_to_update_without_snapshot_suffix` if you think it's better, but I don't know if it's worth the inconsistency.

PR #6, opened 2022-07-21
Add replace_text_in_files action to plugin

tonidero:
https://revenuecats.atlassian.net/browse/CSDK-310

This PR adds a new action to the plugin `replace_text_in_files`. This can be used by the SDKs to automate things like replacing API keys in the project or similar tasks. Each SDK would need to implement a lane that uses this action.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ A plugin including commonly used automation logic for RevenueCat SDKs.
 - `replace_version_number`: This action asks for a new version number and updates all occurences of the old version number (passed as a parameter) in the list of files to update (also passed as a parameter).
 - `create_next_snapshot_version`: This action creates bumps the version to the next minor with a `-SNAPSHOT` suffix and creates a PR with the changes.
 - `create_github_release`: This action will create a github release with the given version number as name and tag and the contents of the CHANGELOG.latest.md file as description. It can also upload files to the release if needed.
+- `replace_text_in_files`: This action will replace all the occurences of the old text with new text in the list of given paths to files.
 
 ## Example
 
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -177,7 +177,7 @@ module Fastlane
         )
       end
 
-      private_class_method def self.replace_in(previous_text, new_text, path, allow_empty: false)
+      def self.replace_in(previous_text, new_text, path, allow_empty: false)
         if new_text.to_s.strip.empty? && !allow_empty
           UI.user_error!("Missing `new_text` in call to `replace_in`, looking for replacement for #{previous_text} 😵.")
         end
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -183,7 +183,7 @@ module Fastlane
         end
         sed_regex = "s|#{previous_text.sub('.', '\\.')}|#{new_text}|"
         backup_extension = '.bck'
-        Actions.sh("sed", '-i', backup_extension, sed_regex, path)
+        Actions.sh("sed", "-i#{backup_extension}", sed_regex, path)
       end
 
       private_class_method def self.ensure_new_branch_local_remote(new_branch)
--- a/spec/helper/revenuecat_internal_helper_spec.rb
+++ b/spec/helper/revenuecat_internal_helper_spec.rb
@@ -408,4 +408,25 @@ describe Fastlane::Helper::RevenuecatInternalHelper do
       )
     end
   end
+
+  describe '.replace_in' do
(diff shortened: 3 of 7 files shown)


taquitos (approved):
🎈🐐

README.md
```diff
@@ -20,6 +20,7 @@ A plugin including commonly used automation logic for RevenueCat SDKs.
 - `replace_version_number`: This action asks for a new version number and updates all occurences of the old version number (passed as a parameter) in the list of files to update (also passed as a parameter).
 - `create_next_snapshot_version`: This action creates bumps the version to the next minor with a `-SNAPSHOT` suffix and creates a PR with the changes.
 - `create_github_release`: This action will create a github release with the given version number as name and tag and the contents of the CHANGELOG.latest.md file as description. It can also upload files to the release if needed.
+- `replace_text_in_files`: This action will replace all the occurences of the old text with new text in the list of given paths to files.
```
NachoSoto, line 23: It's awesome that these are here

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -408,4 +408,25 @@ def setup_stubs
       )
     end
   end
+
+  describe '.replace_in' do
```
NachoSoto, line 412: Now that we have tests for this, could you add one that would cover what I fixed in https://github.com/RevenueCat/purchases-ios/pull/1536?
Example:
- Input: `55413C0025B778E200ECCA5A` 
- Replace: `4.1.3`
- Replacement: `4.4.0-SNAPSHOT`

I see that you copied over the correct implementation with the escaped `.`, so this shouldn't replace anything :)

tonidero: So tests don't run fastlane `sh` actions by default so I couldn't actually test the command itself. But I thought there was value in actually testing the command in this case (pretty simple command). So I made `sh` actions run only for these tests by setting the environment variables `ENV["FORCE_SH_DURING_TESTS"]`. I wouldn't like that for production code but I think it's ok for tests, but lmk what you think!

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -177,13 +177,13 @@ def self.create_github_release(release_version, release_description, upload_asse
...
 
-      private_class_method def self.replace_in(previous_text, new_text, path, allow_empty: false)
+      def self.replace_in(previous_text, new_text, path, allow_empty: false)
         if new_text.to_s.strip.empty? && !allow_empty
           UI.user_error!("Missing `new_text` in call to `replace_in`, looking for replacement for #{previous_text} 😵.")
         end
         sed_regex = "s|#{previous_text.sub('.', '\\.')}|#{new_text}|"
         backup_extension = '.bck'
-        Actions.sh("sed", '-i', backup_extension, sed_regex, path)
+        Actions.sh("sed", "-i#{backup_extension}", sed_regex, path)
```
tonidero, line 186: TIL that `sed` works differently on macosx and linux... I thought about writing code for this instead of doing it using a command, but thought it would be much more inefficient, specially for large files. In the end, I found this: https://stackoverflow.com/a/22084103. I'm not a terminal guru, but I tested it in my machine and in the circleci machines and it worked perfectly fine. Please take a look to make sure this is ok!

aboedo: honestly we might just want to go full ruby rather than `sed`, which will also prevent OS differences

tonidero: Yeah... I was worried about performance in some of those `pbxproj` iOS files... But I think it should probably be fine... I might just run some tests tomorrow and make sure but would probably be cleaner than this.

tonidero: I moved this extra implementation to https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/7 since this PR was already approved and I didn't want to change implementation further. I checked performance and didn't run into issues even when trying to substitute somewhat large files.

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -408,4 +408,53 @@ def setup_stubs
...
+    end
+
+    it 'does not change any text if old text not present in file' do
+      contents = 'Testing 4.1.4'
+      File.write(tmp_test_file_path, contents)
+      Fastlane::Helper::RevenuecatInternalHelper.replace_in('4.1.3', '4.4.0-SNAPSHOT', tmp_test_file_path)
+      expect(File.read(tmp_test_file_path)).to eq(contents)
+    end
+
+    it 'does not change text that does not match exact old text with dots' do
```
NachoSoto, line 453: Awesome 👍🏻

spec/helper/revenuecat_internal_helper_spec.rb
NachoSoto: I think `FORCE_SH_DURING_TESTS` for this subgroup of tests makes sense.

NachoSoto (approved):

PR #7, opened 2022-07-27
Improve replace_in functionality so it's not platform dependant and make it more legible

tonidero:
During the previous PR (#6 ) I noticed that the `sed` is platform dependent (it works differently in macosx and linux) which caused issues with circleci. This PR changes the implementation to use a ruby implementation instead of a command-line based implementation, which should help with legibility and make it platform agnostic.

### DIFF
--- a/fastlane-plugin-revenuecat_internal.gemspec
+++ b/fastlane-plugin-revenuecat_internal.gemspec
@@ -30,7 +30,9 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency('rspec_junit_formatter')
   spec.add_development_dependency('rubocop', '1.31.1')
   spec.add_development_dependency('rubocop-performance')
+  spec.add_development_dependency('rubocop-rake')
   spec.add_development_dependency('rubocop-require_tools')
+  spec.add_development_dependency('rubocop-rspec')
   spec.add_development_dependency('simplecov')
   spec.metadata['rubygems_mfa_required'] = 'true'
 end
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -183,7 +183,6 @@ module Fastlane
         end
         original_text = File.read(path)
         replaced_text = original_text.gsub(previous_text, new_text)
-        File.write("#{path}.bck", original_text)
         File.write(path, replaced_text)
       end
 
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -181,9 +181,10 @@ module Fastlane
         if new_text.to_s.strip.empty? && !allow_empty
           UI.user_error!("Missing `new_text` in call to `replace_in`, looking for replacement for #{previous_text} 😵.")
         end
-        sed_regex = "s|#{previous_text.sub('.', '\\.')}|#{new_text}|"
-        backup_extension = '.bck'
-        Actions.sh("sed", "-i#{backup_extension}", sed_regex, path)
+        original_text = File.read(path)
+        replaced_text = original_text.gsub(previous_text, new_text)
+        File.write("#{path}.bck", original_text)
+        File.write(path, replaced_text)
       end
 
       private_class_method def self.ensure_new_branch_local_remote(new_branch)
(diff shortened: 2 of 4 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -181,9 +181,10 @@ def self.replace_in(previous_text, new_text, path, allow_empty: false)
         if new_text.to_s.strip.empty? && !allow_empty
           UI.user_error!("Missing `new_text` in call to `replace_in`, looking for replacement for #{previous_text} 😵.")
         end
-        sed_regex = "s|#{previous_text.sub('.', '\\.')}|#{new_text}|"
-        backup_extension = '.bck'
-        Actions.sh("sed", "-i#{backup_extension}", sed_regex, path)
+        original_text = File.read(path)
+        replaced_text = original_text.gsub(previous_text, new_text)
+        File.write("#{path}.bck", original_text)
```
tonidero, line 186: Honestly, I'm not sure if we need the backup file having github... But I thought I would keep the same functionality we had with `sed` just in case

spec/actions/replace_version_number_action_spec.rb
```diff
@@ -1,67 +1,15 @@
...
-    it 'replaces old version with new version in passed files when files to update without prerelease modifiers is empty' do
-      expect(Fastlane::Actions).to receive(:sh).with('sed', '-i.bck', 's|1\\.12.0|1.13.0-SNAPSHOT|', './test_file.sh').once
-      expect(Fastlane::Actions).to receive(:sh).with('sed', '-i.bck', 's|1\\.12.0|1.13.0-SNAPSHOT|', './test_file2.rb').once
-      Fastlane::Actions::ReplaceVersionNumberAction.run(
-        current_version: '1.12.0',
-        new_version_number: '1.13.0-SNAPSHOT',
-        files_to_update: ['./test_file.sh', './test_file2.rb'],
-        files_to_update_without_prerelease_modifiers: []
-      )
-    end
```
tonidero, line 64: I replaced all these tests with just making sure it's calling the appropriate helper method. The helper functions are already tested more thoroughly in their own tests.

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -443,6 +466,12 @@ def setup_stubs
       expect(File.read(tmp_test_file_path)).to eq('Testing changing text=4.1.3')
     end
 
+    it 'changes multiple occurences of old string with new string' do
```
tonidero, line 469: Note that this is new. This didn't really happened with the `sed` command where only the first occurence of a string got replaced. I think this is good but lmk if you think otherwise.

fastlane-plugin-revenuecat_internal.gemspec
```diff
@@ -30,7 +30,9 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency('rspec_junit_formatter')
   spec.add_development_dependency('rubocop', '1.31.1')
   spec.add_development_dependency('rubocop-performance')
+  spec.add_development_dependency('rubocop-rake')
   spec.add_development_dependency('rubocop-require_tools')
+  spec.add_development_dependency('rubocop-rspec')
```
tonidero, line 35: Added a couple of extra rubocop dependencies since it was being suggested and didn't cause any harm.

aboedo: 👍

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
aboedo: let's remove the `.bck`, it's never been particularly helpful

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -1,50 +1,75 @@
...
         '1.11.0',
         '1.12.0',
         [file_to_update_1, file_to_update_2],
         [file_to_update_without_prerelease_modifiers_3, file_to_update_without_prerelease_modifiers_4]
       )
+
+      expect(File.read(file_to_update_1)).to eq('Contains version: 1.12.0')
+      expect(File.read(file_to_update_2)).to eq('Contains version: 1.12.0 and other version: 1.11.1')
+      expect(File.read(file_to_update_without_prerelease_modifiers_3)).to eq('Contains version: 1.12.0')
+      expect(File.read(file_to_update_without_prerelease_modifiers_4)).to eq('Contains version: 1.12.0')
```
aboedo, line 34: 👏

spec/helper/revenuecat_internal_helper_spec.rb
aboedo: yeah, probably safer to update all instances

aboedo (approved):
looks good, although I think we can get rid of the `.bck` now

joshdholtz (approved):
😻

NachoSoto (approved):

PR #8, opened 2022-08-01
Create commit_current_changes action

tonidero:
Created new `commit_current_changes` action so it can be used by SDKs.
Also modified `replace_text_in_files` so it supports another parameter indicating whether new text can be empty or not.


NachoSoto (approved):
Nice

PR #9, opened 2022-08-03
Add get_latest_github_release_within_same_major action

tonidero:
This PR moves the action that was added to PHC here: https://github.com/RevenueCat/purchases-hybrid-common/pull/178 to the internal plugin, so it can be shared across all the hybrid SDKs.


lib/fastlane/plugin/revenuecat_internal/actions/get_latest_github_release_within_same_major_action.rb
```diff
@@ -0,0 +1,80 @@
...
+require_relative '../helper/revenuecat_internal_helper'
+
+module Fastlane
+  module Actions
+    class GetLatestGithubReleaseWithinSameMajorAction < Action
+      def self.run(params)
+        repo_name = params[:repo_name]
+        current_version = Gem::Version.new(params[:current_version])
+        UI.message("Getting latest release for #{repo_name} on GitHub within same major as #{current_version}")
+        tag_names = Helper::RevenuecatInternalHelper.get_github_release_tag_names(repo_name)
```
tonidero, line 13: @vegaro I modified this a bit in a few places. Some of the changes you might notice:
- Now we return the list of non-prerelease tag names from the helper and we do the selection here. 
- I also added an extra check to make sure there exists at least a version within the same major.
- I modified the environment variables used to pass parameters to the action to match the namespaces we use in the plugin.
- I added `type` and `optional` parameters to the action parameters, so they are consistent with the other actions in the plugin.
Do lmk if you have any questions or comments!

spec/actions/get_latest_github_release_witin_same_major_action_spec.rb
```diff
@@ -0,0 +1,48 @@
+describe Fastlane::Actions::GetLatestGithubReleaseWithinSameMajorAction do
+  describe '#run' do
+    let(:repo_name) { 'purchases-ios' }
+    let(:get_releases_purchases_ios_response) do
+      { json: JSON.parse(File.read("#{File.dirname(__FILE__)}/../test_files/get_releases_purchases_ios.json")) }
```
tonidero, line 5: I added a minified response from the github API for the tests so they are more thorough.

lib/fastlane/plugin/revenuecat_internal/actions/get_latest_github_release_within_same_major_action.rb
vegaro: Makes sense!

lib/fastlane/plugin/revenuecat_internal/actions/get_latest_github_release_within_same_major_action.rb
```diff
@@ -0,0 +1,80 @@
...
+          FastlaneCore::ConfigItem.new(key: :current_version,
+                                       env_name: "RC_INTERNAL_CURRENT_VERSION",
+                                       type: String,
+                                       optional: false,
+                                       description: "The current release version to check for next minor or patch")
+        ]
+      end
+
+      def self.authors
+        ["vegaro"]
```
vegaro, line 61: add yourself here :D

vegaro (approved):

aboedo (approved):

PR #10, opened 2022-08-04
Splits changelog in sections depending on label

vegaro:
This was originally developed in https://github.com/RevenueCat/purchases-ios/pull/1769 but now we have tests and we can share across all repos 👏

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -41,6 +41,9 @@ module Fastlane
         body = JSON.parse(resp[:body])
         commits = body["commits"].reverse
 
+        supported_types = ["breaking", "build", "ci", "docs", "feat", "fix", "perf", "refactor", "style", "test"]
+        changelog_sections = { :breaking_changes => [], :fixes => [], :new_features => [], :other => [] }
+
         commits.map do |commit|
           if rate_limit_sleep > 0
             UI.message("Sleeping #{rate_limit_sleep} second(s) to avoid rate limit 🐌")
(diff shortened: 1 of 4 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -41,16 +41,16 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
         body = JSON.parse(resp[:body])
         commits = body["commits"].reverse
 
+        supported_types = ["breaking", "build", "ci", "docs", "feat", "fix", "perf", "refactor", "style", "test"]
```
aboedo, line 44: suuuuper minor, FFTI since it won't make any tangible difference in practice:
this _could_ be a set by adding `to_set` at the end, i.e.: 
```suggestion
        supported_types = ["breaking", "build", "ci", "docs", "feat", "fix", "perf", "refactor", "style", "test"].to_set
```

which would make the select on line `70` run _slightly_ faster.

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -61,17 +61,25 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
...
           items = body["items"]
-
           if items.size == 1
             item = items.first
             message = "#{item['title']} (##{item['number']})"
             username = item["user"]["login"]
+            types_of_change = item["labels"]
+                              .map { |label_info| label_info["name"] }
+                              .select { |label| supported_types.include?(label) }
+                              .uniq
```
aboedo, line 71: same very slight optimization here, replacing `.uniq` with `to_set` will also make ever-so-slightly faster to check inclusion in `get_section_depending_on_types_of_change`
```suggestion
            types_of_change = item["labels"]
                              .map { |label_info| label_info["name"] }
                              .select { |label| supported_types.include?(label) }
                              .to_set
```

aboedo (approved):
🚢🚢🚢🚢🚢
I love that this is now common to all SDKs _and_ tested!! 🎉

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -225,6 +233,36 @@ def self.get_github_release_tag_names(repo_name)
...
+
+      private_class_method def self.get_section_depending_on_types_of_change(change_types)
+        if change_types.include?("breaking")
+          :breaking_changes
+        elsif change_types.include?("feat")
+          :new_features
+        elsif change_types.include?("fix")
+          :fixes
+        else
+          :other
```
tonidero, line 263: I guess for SDKs where we are using this but not forcing the labels, most changes will be in `other`. I think it's ok though since we do offer the chance for devs to modify the generated changelog

spec/test_files/get_commit_sha_0e67cdb1c7582ce3e2fd00367acc24db6242c6d6.json
```diff
@@ -33,7 +33,17 @@
                 "type": "User",
                 "site_admin": false
             },
-            "labels": [],
+            "labels": [
```
tonidero, line 36: Maybe have one test where `labels` is empty? Since for other SDKs that use this but don't put labels that will be the common case.

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
vegaro: yes, we will use labels everywhere at some point,  but until them we will probably have to edit the changelogs manually

PR #12, opened 2022-08-08
Not validate current branch when creating next snapshot

tonidero:
This should help dealing with: https://revenuecats.atlassian.net/browse/SDKONCALL-84

With the plugin, I added a validation for the current branch to the action that creates the next prerelease version PR. This validation didn't exist before, and caused the issue that we see in this ticket. This PR removes that validation from that action. After this is merged, I will update the SDKs to use the new version of the plugin


taquitos (approved):
Noice. 🎈🐐

PR #13, opened 2022-08-08
Add update_hybrids_versions_file action

tonidero:
This action will allow to update the `VERSIONS.md` file that exists on the hybrid SDKs with the iOS, Android and PHC versions automatically.


vegaro (approved):
This is amazing!

PR #15, opened 2022-08-09
Calculate next version using labels

vegaro:
Adds a new action that determines the next version depending on the labels of the PRs since the last release.

I also refactored the auto_generate_changelog because some code is shared between them. I added a memory cache so the  responses from GitHub are stored and reused.

[CSDK-380]

[CSDK-380]: https://revenuecats.atlassian.net/browse/CSDK-380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/test_files/get_commits_since_last_release_patch_changes.json
```diff
@@ -0,0 +1,446 @@
+{
```
vegaro, line 1: I added some new responses because I wanted to simulate a patch release, so I got a bunch of commits that represented a patch. I didn't want to make it up so I got a real response

spec/test_files/get_commit_sha_cfdd80f73d8c91121313d72227b4cbe283b57c1e.json
```diff
@@ -33,7 +33,17 @@
                 "type": "User",
                 "site_admin": false
             },
-            "labels": [],
+            "labels": [
```
vegaro, line 36: We now have a `next_release` label so I added it to this commit. We have another commit that doesn't have any labels and we are testing for it

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -9,9 +9,18 @@
 
 module Fastlane
   UI = FastlaneCore::UI unless Fastlane.const_defined?(:UI)
+  SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf refactor style test].to_set
 
   module Helper
     class RevenuecatInternalHelper
+      @cached_commits_by_old_version = {}
+      @pr_resp_items_by_sha = {}
```
tonidero, line 17: Hmm so these are instance variables. Considering all methods in this helper are class methods (basically `static`), I think these should be class variables as well. To do that you should add 2 `@`, Like:
```suggestion
      @@cached_commits_by_old_version = {}
      @@pr_resp_items_by_sha = {}
```
It does work with this according to this answer: https://stackoverflow.com/a/9311665. But I feel these would be better like that.

A different argument would be to change these to not be class methods and create instances of the helper inside the actions... I followed this convention since it's how it's done in fastlane afaik, so I think it's better to leave as is.

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -24,51 +33,59 @@ def self.replace_version_number(previous_version_number, new_version_number, fil
         end
       end
 
+      def self.determine_next_version_using_labels(repo_name, github_token, rate_limit_sleep)
+        old_version = Actions.sh("git describe --tags --abbrev=0").strip
```
tonidero, line 37: Not needed for this PR but one thing I've noticed is that sometimes I haven't fetched the latest tags from origin. Maybe we could do that before this. Again, not needed for this PR and I can do it later as well.

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -24,51 +33,59 @@ def self.replace_version_number(previous_version_number, new_version_number, fil
...
+        UI.important("Determining next version after #{old_version}")
+        org = "RevenueCat"
```
tonidero, line 39: I've been hardcoding the `org` in all the new code I've been writing... I guess it doesn't hurt to make it flexible, but I don't think we will be using this in other org's repos and we avoid having an extra parameter around.

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -24,51 +33,59 @@ def self.replace_version_number(previous_version_number, new_version_number, fil
...
+
+        commits = get_commits_since_old_version(org, github_token, old_version, repo_name)
```
tonidero, line 41: Hmm I'm wondering if we should move all of this code to a different helper, since `RevenueCatInternalHelper` is growing pretty big... Maybe a `ChangelogHelper`, or `GithubHelper`?  I created a second helper for the `update_hybrids_versions_file` action, in case that helps.

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -263,6 +274,85 @@ def self.get_github_release_tag_names(repo_name)
...
+
+        # Get pull request associate with commit message
+        pr_resp = Actions::GithubApiAction.run(server_url: 'https://api.github.com',
+                                               path: "/search/issues?q=repo:#{org}/#{repo_name}+is:pr+base:main+SHA:#{sha}",
+                                               http_method: 'GET',
+                                               body: {},
+                                               api_token: github_token)
+        body = JSON.parse(pr_resp[:body])
+        items = body["items"]
+        @pr_resp_items_by_sha[sha] ||= items
```
tonidero, line 328: Hmm I guess this could be a simple assignment, since we have already checked before that the map doesn't contain the value. NABD though
```suggestion
        @pr_resp_items_by_sha[sha] = items
```

spec/actions/determine_next_version_using_labels_action_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    let(:mock_github_token) { 'mock-github-token' }
+    let(:new_version) { '1.13.0' }
+
+    it 'calls all the appropriate methods with appropriate parameters' do
+      expect(Fastlane::Helper::RevenuecatInternalHelper).to receive(:determine_next_version_using_labels)
+        .with(mock_repo_name, mock_github_token, 3)
+        .and_return(new_version)
+        .once
+
+      Fastlane::Actions::DetermineNextVersionUsingLabelsAction.run(
```
tonidero, line 13: You should assert that the value returned by this is `new_version`. You can store this in a variable and do `expect(calculated_version).to eq(new_version)`

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -74,6 +74,9 @@
   end
 
   describe '.auto_generate_changelog' do
+    before(:each) do
+      Fastlane::Helper::RevenuecatInternalHelper.cleanup_caches
+    end
```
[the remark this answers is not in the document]
tonidero: Yeah... I don't know... Honestly, using class methods is not my preferred choice, but since everything in fastlane seems to be done with them, I followed the convention. I think this works for now...

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -593,4 +580,173 @@ def setup_stubs
...
+      expect do
+        Fastlane::Helper::RevenuecatInternalHelper.determine_next_version_using_labels(
+          'mock-repo-name',
+          'mock-github-token',
+          0
+        )
+      end.to raise_exception(StandardError)
+    end
+
+    def setup_stubs
```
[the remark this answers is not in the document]
tonidero: Hmm the link doesn't seem to direct to the correct place for me, could you post it again? I think I know what code you're mentioning and I agree it would be ideal to extract it... Maybe we can pass a map of sha's to responses as a parameter to generate this programmatically?

Also, maybe we should rename this to `setup_commit_search_stubs` or something like that? So it's not so generic (I should have noticed that when I did the other tests 😅 )

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -209,6 +196,88 @@
                               "* added a log when `autoSyncPurchases` is disabled (#1749) via aboedo (@aboedo)")
     end
 
+    it 'commits response is cached' do
+      setup_stubs
+      cached_commits_by_old_version = Fastlane::Helper::RevenuecatInternalHelper.instance_variable_get(:@cached_commits_by_old_version)
```
tonidero, line 201: Hmm I wouldn't consider this as part of the api of this class so I don't think we should access it. I think we should test this more as part of it's effects, which you're already doing. I mean, when calling the method twice, it should only execute the api calls once. You're already doing that so I think we should just remove these checks IMO.

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -593,4 +580,173 @@ def setup_stubs
       expect(tag_names).to include('5.10.0')
     end
   end
+
+  describe '.determine_next_version_using_labels' do
+    let(:repo_name) { 'purchases-ios' }
+    let(:server_url) { 'https://api.github.com' }
+    let(:http_method) { 'GET' }
+    let(:get_commits_response) do
```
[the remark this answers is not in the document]
tonidero: Hmm I'm not sure whether to extract them or not... If they did nothing (like `repo_name` or `server_url`), I would say to leave them inside since they don't have any logic and it doesn't hurt. These do have a bit of logic which would be good to avoid repetition of, but I also like having them close to the context of the tests... I'm thinking about creating a `read_test_file` method where we can just pass the name of the file and try to minify the repetition. What do you think? 

Also, the rename makes sense 👍

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -263,6 +274,85 @@ def self.get_github_release_tag_names(repo_name)
...
+        return items
+      end
+
+      private_class_method def self.get_commits_since_old_version(org, github_token, old_version, repo_name)
+        return @cached_commits_by_old_version[old_version] if @cached_commits_by_old_version.include?(old_version)
```
tonidero, line 333: Hmm considering we are only doing this 1-2 times and not in a per-commit basis, I'm not sure if it's worth to cache these as well... I guess it doesn't hurt though, so I'm ok with it.

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -9,9 +9,18 @@
...
 module Fastlane
   UI = FastlaneCore::UI unless Fastlane.const_defined?(:UI)
+  SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf refactor style test].to_set
 
   module Helper
     class RevenuecatInternalHelper
+      @cached_commits_by_old_version = {}
+      @pr_resp_items_by_sha = {}
+
+      def self.cleanup_caches
```
tonidero, line 19: Maybe rename to `cleanup_github_commit_caches`?

tonidero (commented):
Great job on this! As we talked, I'm a bit unsure on whether caching the responses is worth the extra complexity... But I think what you have now is not bad, so I think we can go ahead with this. Just some comments

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
vegaro: I did them class instance vars because rubocop and RubyMine complains if I make them class vars:

```
 Style/ClassVars: Replace class var @@pr_resp_items_by_sha with a class instance var.
 ```
 
 So I figured that's what I should be doing. Do you think I should still do class vars?

File: lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb (hunk 13 above)
Comment (vegaro) [reply lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb:333]: I am ok with removing it if you think it's overengineering. Less source of bugs

File: spec/helper/revenuecat_internal_helper_spec.rb (hunk 10 above)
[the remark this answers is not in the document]
Comment (vegaro) [reply spec/helper/revenuecat_internal_helper_spec.rb:824]: I am talking about the `setup_stubs` function in `auto_generate_changelog`

File: lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb (hunk 3 above)
Comment (tonidero) [reply lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb:17]: Hmm interesting... https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ClassVars
I trust the default rubocop rules, and considering it works like this i think we should leave it as you have it, but I don't know if this explanation warranted disabling a whole language feature... 🤷‍♂️

File: lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb (hunk 13 above)
Comment (tonidero) [reply lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb:333]: Hmm I'm going to vote for removal... I think we don't need to early optimize this just yet, but not a strong opinion.

**Comment by tonidero** [comment]
> @tonidero I added a hashes_to_responses parameter in [8eb7640](https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/commit/8eb76407f659c9ac4b90220122349d52c544e39a), but the tests are failing because:
> 
> > `get_commit_1_response` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).
> 
> Do you know how to fix it?

Ahh you will need to declare the map with a `let` I think. Basically the variables declared with `let` are only accesible inside a test or another let I think, since they are recalculated on each test. Could you try that?

**Comment by vegaro** [comment]
@tonidero that worked! Thanks!

File: lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb [hunk 15]
```diff
@@ -24,64 +25,6 @@ def self.replace_version_number(previous_version_number, new_version_number, fil
...
-
-            line = "* #{message} via #{name} (@#{username})"
-            changelog_sections[section].push(line)
-          else
-            UI.user_error!("Cannot generate changelog. Multiple commits found for #{sha}")
-          end
-        end
-
-        build_changelog_sections(changelog_sections)
-      end
```
vegaro, line 83: Extracted to VersioningHelper

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -233,36 +171,6 @@ def self.get_github_release_tag_names(repo_name)
...
-          when :fixes
-            title = "### Bugfixes"
-          when :new_features
-            title = "### New Features"
-          when :other
-            title = "### Other Changes"
-          end
-          "#{title}\n#{prs.join("\n")}"
-        end.join("\n")
-      end
```
vegaro, line 253: Extracted to VersioningHelper

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -233,36 +171,6 @@ def self.get_github_release_tag_names(repo_name)
...
-        if change_types.include?("breaking")
-          :breaking_changes
-        elsif change_types.include?("feat")
-          :new_features
-        elsif change_types.include?("fix")
-          :fixes
-        else
-          :other
-        end
-      end
```
vegaro, line 265: Extracted to VersioningHelper

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -0,0 +1,145 @@
...
+        if change_types.include?("breaking")
+          :major
+        elsif change_types.include?("feat")
+          :minor
+        else
+          :patch
+        end
+      end
+
+      def self.increase_version(current_version, type_of_bump, snapshot)
```
tonidero, line 113: Nitpick, I would move public methods above private ones.

tonidero: Also, I believe we need to add tests for this function, even though it's already tested through some of the usages. I would keep a test for those usages and test it more thoroughly in the tests for the helper.

tonidero (approved):
Just a couple of comments but looks good!

spec/helper/versioning_helper_spec.rb
```diff
@@ -0,0 +1,305 @@
+describe Fastlane::Helper::RevenuecatInternalHelper do
```
tonidero, line 1: Small thing
```suggestion
describe Fastlane::Helper::VersioningHelper do
```

vegaro: oopsies

tonidero (approved):

PR #16, opened 2022-08-09
`README`: fixed command for running tests

NachoSoto:
`rake` fails for me, but this is the command that `.circleci/config.yml` runs, and it works.


tonidero (approved):
That makes sense. It works for me with only `rake` but it should be better to use `bundle exec rake`.

PR #19, opened 2022-08-15
Use latest tag from repo instead of from branch

tonidero:
This PR updates our code to generate the changelog and calculate the next version automatically to use the latest non-prerelease tag in the repo instead of the latest tag in the current branch. This is because we changed our process to perform the release tags from a branch, instead of from `main`. This caused the previous method to not pickup these tags, since the release branch is squashed in the merge so there is no access from main to the branch.


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -39,7 +39,8 @@ def self.determine_next_version_using_labels(repo_name, github_token, rate_limit
       end
 
       def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
-        old_version = Actions.sh("git describe --tags --abbrev=0").strip
+        Actions.sh("git fetch --tags")
```
tonidero, line 42: Added this since a couple of times now, I've tried to bump but forgot to fetch latest tags first.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -94,6 +95,12 @@ def self.increase_version(current_version, type_of_bump, snapshot)
         end
       end
 
+      private_class_method def self.latest_non_prerelease_version_number
+        Actions
+          .sh("git describe --tags $(git rev-list --exclude=\"*[a-zA-Z]*\" --tags=\"[0-9]*.[0-9]*.*[0-9]\" --max-count=1)")
```
tonidero, line 100: This is a bit messy... But the `exclude` and `tags` options in the `git rev-list` command accept a `glob` and not a `regex` which was making filtering out unwanted tags difficult. In the end, the current behavior will be that it will filter out tags containing any letters, and include tags that match this glob `[0-9]*.[0-9]*.*[0-9]`. Lmk if you know of a better way to do this!

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
tonidero: Note that I didn't add this in the `determine_next_version_using_labels` since that method is currently only intended to be used in automated methods so it seemed unnecessary

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: I think the only alternative would be to do it programmatically using ruby and calculating the highest tag?

vegaro (approved):

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
tonidero: Yeah that's the best alternative... I wasn't sure about doing that since the list of tags can grow... But honestly, I think it's not going to be a problem. And I think it's going to be cleaner so I think I will do that now.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -94,6 +95,15 @@ def self.increase_version(current_version, type_of_bump, snapshot)
         end
       end
 
+      private_class_method def self.latest_non_prerelease_version_number
+        Actions
+          .sh("git tag", log: false)
```
tonidero, line 100: Added the `log: false` to not output all the tags in the console while doing this, since there can be a lot

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
tonidero: I did the change. Lmk what you think @vegaro

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: beautiful

vegaro (approved):

PR #20, opened 2022-08-16
Validate branch with prompt

tonidero:
This PR removes the automatic branch validation to `main` we do during the `bump_version_update_changelog_create_pr` action. Instead, we will prompt the user to confirm that the current branch is the one intended to be the base of the bump.


vegaro (approved):
@joshdholtz got hit by this yesterday so this is a good change

PR #22, opened 2022-08-17
Returns the type of bump when determining next version

vegaro:
We need to skip automatic releases if the next release is major, so I added the type of bump as a return for the action

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
@@ -35,7 +35,7 @@ module Fastlane
           end
         end
         UI.important("Type of bump after version #{old_version} is #{type_of_bump}")
-        increase_version(old_version, type_of_bump, false)
+        return increase_version(old_version, type_of_bump, false), type_of_bump
       end
 
       def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
--- a/lib/fastlane/plugin/revenuecat_internal/actions/determine_next_version_using_labels_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/determine_next_version_using_labels_action.rb
@@ -15,7 +15,7 @@ module Fastlane
       end
 
       def self.description
-        "Determines next version using the labels of the pull requests merged since the last release"
+        "Determines next version using the labels of the pull requests merged since the last release. Returns the next version number along with the type of bump."
       end
 
       def self.authors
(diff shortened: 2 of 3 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -35,7 +35,7 @@ def self.determine_next_version_using_labels(repo_name, github_token, rate_limit
           end
         end
         UI.important("Type of bump after version #{old_version} is #{type_of_bump}")
-        increase_version(old_version, type_of_bump, false)
+        return increase_version(old_version, type_of_bump, false), type_of_bump
```
vegaro, line 38: Not sure which is better, if this or returning a hash

tonidero: Hmm honestly, I've never been much of a fan of returning multiple values for the same function... I would prefer to return an object... What do you think about separating the action into 2, one to get the type of bump from the changelog, and another to calculate the next version? It's true that it will slightly more logic to the sdks, but I feel it would be cleaner

tonidero (commented):
The change itself should work but I wanted to discuss alternatives, since I'm not a big fan of returning multiple values in the same function.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: Do you mind if I go with the returning of an object for now?

vegaro: Ok, I think I can just make this function return the type of change and expose increase version as an action

tonidero: 👍 Thanks! I'm ok merging this if we want to get it working fast but I would definitely add a task to tackle this at some point later. Do lmk if so

vegaro: ok, yeah, I was going to suggest that because I wanted to get the release trains going and this should be the last piece of the puzzle (I think). So maybe I change this to return an object for now and create a task to split in 2 actions. Is that cool?

tonidero: After thinking about it, I think it's better to leave as is, since returning an object would mean we would have to expose it from the plugin to the sdks, and not sure if that's a good idea. So I think we can merge this as is and just change to 2 actions later

tonidero (approved):

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: Created a task https://revenuecats.atlassian.net/browse/CSDK-417

PR #23, opened 2022-08-17
Support prerelease version during version increase

tonidero:
This PR adds support for versions with prerelease modifiers when trying to bump versions automatically. Now we will support versions like `1.2.3-alpha.1`, `2.2.2-beta.3` or `3.3.3-rc.4` which will increase to the version without the prerelease modifier and add `SNAPSHOT` if needed. The regex for valid version numbers now is: `^[0-9]+.[0-9]+.[0-9]+(-(alpha|beta|rc).[0-9]+)?$`.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
@@ -72,20 +72,29 @@ module Fastlane
       end
 
       def self.increase_version(current_version, type_of_bump, snapshot)
-        version_split = current_version.split('.')
-        UI.user_error("Invalid version number: #{current_version}. Expected 3 numbers separated by '.'") if version_split.size != 3
+        is_prerelease = %w(alpha beta rc).any? { |prerelease| current_version.include?(prerelease) }
+        is_valid_version = current_version.match?("^[0-9]+.[0-9]+.[0-9]+(-(alpha|beta|rc).[0-9]+)?$")
+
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -72,20 +72,29 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
       end
 
       def self.increase_version(current_version, type_of_bump, snapshot)
-        version_split = current_version.split('.')
-        UI.user_error("Invalid version number: #{current_version}. Expected 3 numbers separated by '.'") if version_split.size != 3
```
tonidero, line 76: Note that I got an error here, it was supposed to be `user_error!`

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -257,9 +257,9 @@
       expect(next_version).to eq('1.12.0-SNAPSHOT')
     end
 
-    it 'calculates next version correctly with snapshot version' do
```
tonidero, line 260: Note that we won't support bumping `SNAPSHOT` versions automatically after this. I did this change since I think we should never find ourselves with a SNAPSHOT. This function is currently used in 2 places:
- When determining the next version automatically for automatic bumps, which uses the latest tag in the repo, which should never be a SNAPSHOT
- When calculating the next prerelease version, and the version in the release branch when we execute this should also never be a SNAPSHOT.

Do lmk if you think this is a problem and I can refine the regex to support snapshots as well.

NachoSoto (approved):

PR #24, opened 2022-08-17
`bump_version_update_changelog_create_pr_action`: validate new version is not empty

NachoSoto:
I realized the script kept going after I just pressed "enter" with an empty version number.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
@@ -29,6 +29,7 @@ module Fastlane
         # Ask for new version number
         new_version_number ||= UI.input("New version number: ")
 
+        UI.user_error!("Version number cannot be empty") if new_version_number.strip.empty?
         UI.important("New version is #{new_version_number}")
 
         new_branch_name = "release/#{new_version_number}"
--- a/spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
+++ b/spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
@@ -17,7 +17,7 @@ describe Fastlane::Actions::BumpVersionUpdateChangelogCreatePrAction do
       allow(FastlaneCore::UI).to receive(:interactive?).and_return(true)
       allow(FastlaneCore::UI).to receive(:confirm).with(anything).and_return(true)
       allow(FastlaneCore::UI).to receive(:input).with('New version number: ').and_return('')
-      
+
       expect(FastlaneCore::UI).to receive(:user_error!)
         .with('Version number cannot be empty')
         .once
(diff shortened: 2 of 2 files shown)


NachoSoto:
Not sure if this is the best way, and also maybe I should add a test for this?

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -29,6 +29,7 @@ def self.run(params)
         # Ask for new version number
         new_version_number ||= UI.input("New version number: ")
 
+        UI.user_error!("Invalid version number: #{new_version_number}") unless !new_version_number.strip.isEmpty?
```
tonidero, line 32: I think something like this? And yeah, maybe writing a test to make sure it fails in this case would be great 😄 
```suggestion
        UI.user_error!("Version number cannot be empty") if new_version_number.blank?
```

NachoSoto: Blank doesn't exist though?

NachoSoto:
@tonidero updated with tests!

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
tonidero: Hmm weird... https://apidock.com/rails/Object/blank%3F it's there... Did you add the `?`?

tonidero:
Looks like there is a small linter error, other than that `.strip.empty?` should be the same as `.blank?` so it should work as well. Not sure why `blank?` wouldn't work though...

spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
```diff
@@ -13,6 +13,29 @@
...
+      allow(FastlaneCore::UI).to receive(:interactive?).and_return(true)
+      allow(FastlaneCore::UI).to receive(:confirm).with(anything).and_return(true)
+      allow(FastlaneCore::UI).to receive(:input).with('New version number: ').and_return('')
+      
+      expect(FastlaneCore::UI).to receive(:user_error!)
+        .with('Version number cannot be empty')
+        .once
+        .and_throw(:expected_error)
+
+      catch :expected_error do
```
tonidero, line 26: I've usually catched errors like this: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/blob/806c2787a95522facb437be5cfad803b74f5e5fe/spec/helper/revenuecat_internal_helper_spec.rb#L100-L102 but this way will also catch the exact error message, so it's more complete 👍

tonidero (approved):

NachoSoto:
I get this:
```
NoMethodError:
       undefined method `blank?' for "1.13.0":String
       Did you mean?  black
                      black!
```

NachoSoto:
Looks like that's part of `active_support`?

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
NachoSoto: That's in `rails`, but we don't use it?

NachoSoto:
That keeps failing to lint but it's passing locally :/ any idea why? I got the failure locally, I removed that trailing whitespace, and now they pass locally, but still fail in CI.

PR #27, opened 2022-08-18
Skip release if no public changes

vegaro:
We are triggering automatic releases in purchases-hybrid-common when merging into main. We shouldn’t trigger a release if the new changes don’t need a release (like maybe changes to ci, or new releases commits)

[CSDK-423]

[CSDK-423]: https://revenuecats.atlassian.net/browse/CSDK-423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/determine_next_version_using_labels_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/determine_next_version_using_labels_action.rb
@@ -10,9 +10,8 @@ module Fastlane
         repo_name = params[:repo_name]
         github_token = params[:github_token]
         rate_limit_sleep = params[:github_rate_limit]
-        skip_if_no_public_changes = params[:skip_if_no_public_changes]
 
-        Helper::VersioningHelper.determine_next_version_using_labels(repo_name, github_token, rate_limit_sleep, skip_if_no_public_changes)
+        Helper::VersioningHelper.determine_next_version_using_labels(repo_name, github_token, rate_limit_sleep)
       end
 
       def self.description
@@ -40,13 +39,7 @@ module Fastlane
                                        description: "Sets a rate limiter for github requests when creating the changelog",
                                        optional: true,
                                        default_value: 0,
-                                       type: Integer),
-          FastlaneCore::ConfigItem.new(key: :skip_if_no_public_changes,
-                                       env_name: "RC_INTERNAL_SKIP_IF_NO_PUBLIC_CHANGES",
-                                       description: "Skip next version if no public changes have been commited to the repo. For example, if the only changes have been changes to CI, there shouldn't be a next release",
-                                       optional: true,
-                                       default_value: false,
-                                       type: Boolean)
+                                       type: Integer)
         ]
       end
 
--- a/lib/fastlane/plugin/revenuecat_internal/actions/determine_next_version_using_labels_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/determine_next_version_using_labels_action.rb
@@ -10,8 +10,9 @@ module Fastlane
         repo_name = params[:repo_name]
         github_token = params[:github_token]
         rate_limit_sleep = params[:github_rate_limit]
+        skip_if_no_public_changes = params[:skip_if_no_public_changes]
 
-        Helper::VersioningHelper.determine_next_version_using_labels(repo_name, github_token, rate_limit_sleep)
+        Helper::VersioningHelper.determine_next_version_using_labels(repo_name, github_token, rate_limit_sleep, skip_if_no_public_changes)
       end
 
       def self.description
@@ -39,7 +40,13 @@ module Fastlane
                                        description: "Sets a rate limiter for github requests when creating the changelog",
                                        optional: true,
                                        default_value: 0,
-                                       type: Integer)
+                                       type: Integer),
+          FastlaneCore::ConfigItem.new(key: :skip_if_no_public_changes,
+                                       env_name: "RC_INTERNAL_SKIP_IF_NO_PUBLIC_CHANGES",
+                                       description: "Skip next version if no public changes have been commited to the repo. For example, if the only changes have been changes to CI, there shouldn't be a next release",
+                                       optional: true,
+                                       default_value: false,
+                                       type: Boolean)
         ]
       end
 
--- a/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
@@ -15,7 +15,8 @@ module Fastlane
         commits = Helper::GitHubHelper.get_commits_since_old_version(github_token, old_version, repo_name)
 
         type_of_bump = :patch
-        should_skip_release = true
+        has_public_changes = false
+        public_change_labels = %w[breaking docs feat fix perf]
         commits.each do |commit|
           break if type_of_bump == :major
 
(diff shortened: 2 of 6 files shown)


vegaro:
@tonidero before I jump into fixing tests, do you think this makes sense?

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -8,35 +8,40 @@ module Fastlane
...
-
-            types_of_change = get_type_of_change_from_pr_info(item)
+          UI.user_error!("Cannot determine next version. Multiple commits found for #{sha}") if items.size != 1
 
-            type_of_bump_for_change = get_type_of_bump_from_types_of_change(types_of_change)
+          item = items.first
+          types_of_change = get_type_of_change_from_pr_info(item)
+          type_of_bump_for_change = get_type_of_bump_from_types_of_change(types_of_change)
+          type_of_bump = type_of_bump_for_change unless type_of_bump_for_change == :patch
+          changes_are_public = (types_of_change & %w[breaking build docs feat fix perf refactor]).size > 0
```
tonidero, line 32: Hmm I wonder if `build` and `refactor` should be considered `public`...

lib/fastlane/plugin/revenuecat_internal/actions/determine_next_version_using_labels_action.rb
```diff
@@ -10,8 +10,9 @@ def self.run(params)
         repo_name = params[:repo_name]
         github_token = params[:github_token]
         rate_limit_sleep = params[:github_rate_limit]
+        skip_if_no_public_changes = params[:skip_if_no_public_changes]
 
-        Helper::VersioningHelper.determine_next_version_using_labels(repo_name, github_token, rate_limit_sleep)
+        Helper::VersioningHelper.determine_next_version_using_labels(repo_name, github_token, rate_limit_sleep, skip_if_no_public_changes)
```
tonidero, line 15: Hmm I wonder, would this ever be `false`? Not sure where we would need this option.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -8,35 +8,40 @@ module Fastlane
...
           break if type_of_bump == :major
 
           sha = commit["sha"]
           items = Helper::GitHubHelper.get_pr_resp_items_for_sha(sha, github_token, rate_limit_sleep, repo_name)
 
-          if items.size == 1
-            item = items.first
-
-            types_of_change = get_type_of_change_from_pr_info(item)
+          UI.user_error!("Cannot determine next version. Multiple commits found for #{sha}") if items.size != 1
```
tonidero, line 26: 👍

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -8,35 +8,40 @@ module Fastlane
...
+      # rubocop:disable Metrics/CyclomaticComplexity
+      def self.determine_next_version_using_labels(repo_name, github_token, rate_limit_sleep, skip_if_no_public_changes)
         old_version = latest_non_prerelease_version_number
         UI.important("Determining next version after #{old_version}")
 
         commits = Helper::GitHubHelper.get_commits_since_old_version(github_token, old_version, repo_name)
 
         type_of_bump = :patch
-
+        should_skip_release = true
```
tonidero, line 19: Hmm I wonder if we should rename this to `has_public_changes = false` and set to true if we find a PR with public changes... I think I find it less confusing and it better represents what this variable contains.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -8,35 +8,40 @@ module Fastlane
...
         end
         UI.important("Type of bump after version #{old_version} is #{type_of_bump}")
+
+        if should_skip_release && skip_if_no_public_changes
+          return old_version, :skip
+        end
+
         return increase_version(old_version, type_of_bump, false), type_of_bump
       end
+      # rubocop:enable Metrics/CyclomaticComplexity
```
tonidero, line 44: Hmm it would be good to refactor this, but didn't have a quick idea to do it so I think we can go with this for now.

tonidero (commented):
Mostly some questions, but looks good.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: good point, they can break stuff so I think they should, I will update

lib/fastlane/plugin/revenuecat_internal/actions/determine_next_version_using_labels_action.rb
vegaro: Good point... I think I am over optimizing here so I will remove the parameter

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: since we talked about splitting it in two (one for the version and another one for the type of bump) I figured the split would remove the warning so I silenced it for now

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: Oh I see, lol. You think they shouldn't. The way I understood this, we would like to release things that would change the code, and potentially generate new bugs.

But now that I think about it, I think we should remove `build` and `refactor` because they are not really public. That would mean less releases and less noise too.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: I could actually remove this by removing the unnecessary parameter from the function

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -15,26 +15,31 @@ def self.determine_next_version_using_labels(repo_name, github_token, rate_limit
         commits = Helper::GitHubHelper.get_commits_since_old_version(github_token, old_version, repo_name)
 
         type_of_bump = :patch
-
+        has_public_changes = false
+        public_change_labels = %w[breaking docs feat fix perf]
```
NachoSoto, line 19: PCH's PRs that update iOS/Android will have `build` (https://github.com/RevenueCat/purchases-hybrid-common/pull/208) as the `label`. In that case we would also want to make a new release.

In general maybe not all `build` PRs though, so I wonder if we should change that to `dependencies`, and include that in this list. What do you think?

NachoSoto (commented):
Nice 👍🏻

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
tonidero: Ah great catch! Yeah, your idea to separate them sounds good to me

vegaro: Great catch. Will update those to be dependencies instead 😄

tonidero (approved):

PR #28, opened 2022-08-19
Fix for changelog not being generated properly

vegaro:
As noted in https://github.com/RevenueCat/purchases-hybrid-common/pull/209, the changelog was not being populated properly because `edit_changelog` is the function that updates the latest changelog file

[CSDK-422]

[CSDK-422]: https://revenuecats.atlassian.net/browse/CSDK-422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -49,7 +49,7 @@ module Fastlane
         end
       end
 
-      def self.prepare_changelog(prepopulated_changelog, changelog_latest_path)
+      def self.write_changelog(prepopulated_changelog, changelog_latest_path)
         UI.user_error!("Pre populated content for changelog was empty") if prepopulated_changelog.empty?
 
         UI.message("Using pre populated contents:\n#{prepopulated_changelog}")
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -49,6 +49,16 @@ module Fastlane
         end
       end
 
+      def self.prepare_changelog(prepopulated_changelog, changelog_latest_path)
+        changelog_filename = File.basename(changelog_latest_path)
+
+        UI.user_error!("Pre populated content for changelog was empty") if prepopulated_changelog.empty?
+
+        UI.message("Using pre populated contents:\n#{prepopulated_changelog}")
+
+        File.write(changelog_latest_path, prepopulated_changelog)
+      end
+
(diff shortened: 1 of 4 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -49,6 +49,15 @@ def self.edit_changelog(prepopulated_changelog, changelog_latest_path, editor)
         end
       end
 
+      def self.write_changelog(prepopulated_changelog, changelog_latest_path)
+        UI.user_error!("Pre populated content for changelog was empty") if prepopulated_changelog.empty?
+
+        UI.message("Using pre populated contents:\n#{prepopulated_changelog}")
```
tonidero, line 55: Hmm I wondered if this would be too long to log, but considering this would be automated, I think that this is useful 👍

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -127,6 +127,27 @@
     end
   end
 
+  describe '.write_changelog' do
+    let(:prepopulated_changelog) { 'mock prepopulated changelog' }
+    let(:changelog_latest_path) { './mock-path/CHANGELOG.latest.md' }
+
+    before(:each) do
+      allow(File).to receive(:write).with(changelog_latest_path, prepopulated_changelog)
```
tonidero, line 135: We could let it actually run the change in a tmp file and check that the file was correctly updated. I've done that in a few tests for extra security... But I think it's ok like this as well.

tonidero (approved):
Just one suggestion to actually test using the file system for completeness, but not really needed. Looks good!

spec/helper/revenuecat_internal_helper_spec.rb
vegaro: I agree, we would have caught the issue with that test

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
vegaro: I think it's useful yes

spec/helper/revenuecat_internal_helper_spec.rb
vegaro: Updated in [d729229](https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/28/commits/d729229b02a735bff288aa369e4db39a7e5d1ac3)

tonidero (approved):

NachoSoto (commented):
🙌🏻

PR #34, opened 2022-09-26
Creates BumpPhcVersionAction

vegaro:
I created this action because its logic was going to be duplicated across all hybrids.

It basically updates the version numbers, and opens a PR, if the `open_pr` parameter is true. I wanted to reuse it for both local and automatic updates, that's why the `open_pr` parameter.

See an example of its usage in https://github.com/RevenueCat/purchases-flutter/pull/473/files#diff-dff4b99d4e651ab9d7597876ec8dbe92aa934e42c0fb55f4aadd6c106fc96688R163

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_create_pr_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_create_pr_action.rb
@@ -1,0 +1,109 @@
+require 'fastlane/action'
+require 'fastlane_core/configuration/config_item'
+require_relative '../helper/revenuecat_internal_helper'
+require_relative '../helper/versioning_helper'
+
+module Fastlane
+  module Actions
+    class BumpPhcVersionCreatePrAction < Action
+      def self.run(params)
+        version_number = params[:current_version]
+        files_to_update = params[:files_to_update]
+        repo_name = params[:repo_name]
+        github_pr_token = params[:github_pr_token]
+        new_version_number = params[:next_version]
+        automatic_release = params[:automatic_release]
+        open_pr = params[:open_pr]
+
+        UI.important("Current version is #{version_number}")
+
+        # Ask for new version number
+        new_version_number ||= UI.input("New version number: ")
+
+        UI.user_error!("Version number cannot be empty") if new_version_number.strip.empty?
+        UI.important("New version is #{new_version_number}")
+
+        if open_pr
+          current_branch = Actions.git_branch
+          if UI.interactive? && !UI.confirm("Current branch is #{current_branch}. Are you sure this is the base branch for your bump?")
+            UI.user_error!("Cancelled during branch confirmation")
+          end
+
+          new_branch_name = "bump-phc/#{new_version_number}"
+
+          Helper::RevenuecatInternalHelper.validate_local_config_status_for_bump(new_branch_name, github_pr_token)
+
+          Helper::RevenuecatInternalHelper.create_new_branch_and_checkout(new_branch_name)
+        end
+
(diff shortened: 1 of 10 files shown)


lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_create_pr_action.rb
```diff
@@ -0,0 +1,109 @@
...
+        ["César de la Vega"]
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :current_version,
+                                       description: "Current version of the SDK",
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :files_to_update,
```
tonidero, line 74: I'm guessing for PHC, we won't have any files that require updating the version number without prerelease modifiers right? Afaik, that's only needed on iOS plist files right now so probably PHC version isn't updated there. (And also, we probably don't want to bump to a prerelease version of PHC except when testing locally)

lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_create_pr_action.rb
```diff
@@ -0,0 +1,109 @@
...
+                                       description: "Github token to use to create the release PR",
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :next_version,
+                                       description: "Next version of the SDK",
+                                       optional: true,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :automatic_release,
+                                       description: "If this is an automatic release",
+                                       optional: true,
```
tonidero, line 95: Since this and `open_pr` are optional, should we add a default value?

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -80,7 +80,7 @@ def self.create_new_branch_and_checkout(branch_name)
         Actions.sh("git checkout -b '#{branch_name}'")
       end
 
-      def self.commmit_changes_and_push_current_branch(commit_message)
+      def self.commit_changes_and_push_current_branch(commit_message)
```
tonidero, line 83: 🔂 😅

lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_create_pr_action.rb
```diff
@@ -0,0 +1,109 @@
+require 'fastlane/action'
+require 'fastlane_core/configuration/config_item'
+require_relative '../helper/revenuecat_internal_helper'
+require_relative '../helper/versioning_helper'
+
+module Fastlane
+  module Actions
+    class BumpPhcVersionCreatePrAction < Action
```
tonidero, line 8: Since creating the PR is an optional operation, should we rename this action to `BumpPhcVersion` only?

tonidero (approved):
LGTM!

lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_create_pr_action.rb
vegaro: Yes, I didn't think we needed it

lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_create_pr_action.rb
vegaro: I think I am going to set it to false, and rename the action as you suggested in the other comment

PR #35, opened 2022-09-30
Sets reviewers for all created PRs

vegaro:
I thought it would be good to set it to `core-sdk` for all created PRs from any automation


NachoSoto:
It would be nice to have a workflow that updates this plugin automatically in all repos 😇

PR #38, opened 2022-10-13
Fix for taking into account commits with no PR

vegaro:


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -69,6 +75,13 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
 
             line = "* #{message} via #{name} (@#{username})"
             changelog_sections[section].push(line)
+          elsif items.size == 0
+            UI.important("Cannot find pull request associated to #{sha}. Using commit information and adding it to the Other section")
+            message = commit["commit"]["message"]
+            name = commit["commit"]["author"]["name"]
+            username = commit["author"]["login"]
+            line = "* #{message} via #{name} (@#{username})"
+            changelog_sections[:other].push(line)
```
aboedo, line 84: let's add a slack warning so we _know_ when this is happening, since we're assuming that this is a patch change but we don't have a way of asserting it

vegaro: Do you mind if I add this in another PR? I think we have RN will also get affected by this bug (I pushed to main by error the other day https://github.com/RevenueCat/react-native-purchases/commit/32309f49903848b432abdfab1bfaf425f42cb76a). We don't have any slack notification from fastlane yet and I worry it will take me time to setup and test

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -153,8 +147,10 @@ def self.increase_version(current_version, type_of_bump, snapshot)
           :major
         elsif change_types.include?("feat")
           :minor
-        else
+        elsif (change_types & PUBLIC_CHANGE_LABELS).size > 0
```
vegaro, line 150: @tonidero this could use the map of all labels and their expected type you suggested in https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/39#discussion_r995758518

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -77,6 +73,8 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
       end
 
       def self.increase_version(current_version, type_of_bump, snapshot)
```
tonidero, line 75: Maybe rename to `calculate_next_version`? Since it won't always increase the version number now.

spec/test_files/get_commit_sha_4ceaceb20e700b92197daf8904f5c4e226625d8a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "total_count": 0,
+  "incomplete_results": false,
+  "items": [
+
+  ]
```
tonidero, line 6: You probably already verified this but just to confirm, this will be the expected result from Github's API if there is no PR associated with the corresponding commit right?

tonidero (commented):
Looking good but will hold approving in case you want to change to use a map to calculate the type of change.

spec/test_files/get_commit_sha_4ceaceb20e700b92197daf8904f5c4e226625d8a.json
vegaro: yes, that's where I got the response from

tonidero (approved):
LGTM!

PR #39, opened 2022-10-14
Add minor label

vegaro:
If a PR is labeled as minor, it should be a minor release. This will happen for example if a dependency is increased in a minor version.


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -151,7 +151,7 @@ def self.increase_version(current_version, type_of_bump, snapshot)
       private_class_method def self.get_type_of_bump_from_types_of_change(change_types)
         if change_types.include?("breaking")
           :major
-        elsif change_types.include?("feat")
+        elsif change_types.include?("feat") || change_types.include?("minor")
```
tonidero, line 154: Should we have an array of `minor_warrant_labels` or something like that? Alternatively we could also have a map of all labels and their expected type of bump to simplify this.

vegaro: Hmmm I actually think I like the map idea. I could use that in https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/38/files#diff-8072699fc7bd400a28a8b259d5ae451c6a5386cc5ff0ac300210ffd16101da66R150

spec/helper/versioning_helper_spec.rb
```diff
@@ -227,6 +230,26 @@
       expect(type_of_bump).to eq(:minor)
     end
 
+    it 'forces a minor update if labeled as minor' do
```
tonidero, line 233: Hmm `forces` seems like a strong word... Maybe `determine next version as minor if labeled as minor`?

tonidero (approved):

PR #40, opened 2022-11-01
Update team reviewers to coresdk

vegaro:
Team name has changed and automatic PRs are failing to add automatic reviewers.

This used to work when the team was `core-sdk` but now it looks like even updating it to `coresdk` fails (at least requesting reviewers using their API). It looks like GitHub might have a bug, related to name changes

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -95,7 +95,7 @@ module Fastlane
           head: head_branch,
           api_url: 'https://api.github.com',
           labels: labels,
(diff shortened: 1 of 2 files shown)


tonidero (approved):
> This used to work when the team was core-sdk but now it looks like even updating it to coresdk fails

Hmm so then, does this work? I'm not clear from that comment. If not should we look for an alternative? 

If it does work, :shipit:

vegaro:
Right now it doesn't work, and the main reason is that the team name is wrong. @aboedo updated it to `coresdk`.

But even when calling the GitHub API to request reviewers using the correct team name, the API responds with an error code. When the name was `core-sdk` the same API would work.

I wanted to update the team name to the correct one as a first step, just in case it's some problem with GitHub that they solve, or in case it works with Fastlane from CI.

PR #41, opened 2022-11-08
Adds minor label if the purchases-hybrid-common bump is a minor

vegaro:
- Adds a `minor` label if the purchases-hybrid-common bump is a minor. This minor label will be detected and will force a minor bump in the next release
- Updates title and body of PR from `Automatic release` to `Automatic bump`

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_action.rb
@@ -23,13 +23,13 @@ module Fastlane
         UI.user_error!("Version number cannot be empty") if new_version_number.strip.empty?
         UI.important("New version is #{new_version_number}")
 
+        new_branch_name = "bump-phc/#{new_version_number}"
+
         if open_pr
           current_branch = Actions.git_branch
           if UI.interactive? && !UI.confirm("Current branch is #{current_branch}. Are you sure this is the base branch for your bump?")
             UI.user_error!("Cancelled during branch confirmation")
           end
-
-          new_branch_name = "bump-phc/#{new_version_number}"
 
           Helper::RevenuecatInternalHelper.validate_local_config_status_for_bump(new_branch_name, github_pr_token)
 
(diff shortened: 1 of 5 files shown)


spec/helper/versioning_helper_spec.rb
```diff
@@ -453,6 +453,23 @@
...
+      bump_type = Fastlane::Helper::VersioningHelper.detect_bump_type('1.2.3', '1.2.4')
+      expect(bump_type).to eq(:patch)
+    end
+
+    it 'correctly detects minor bumps' do
+      bump_type = Fastlane::Helper::VersioningHelper.detect_bump_type('1.2.3', '1.3.0')
+      expect(bump_type).to eq(:minor)
+    end
+
+    it 'correctly detects major bumps' do
```
tonidero, line 467: Could add a test for passing the same versions

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -107,6 +107,25 @@ def self.calculate_next_version(current_version, type_of_bump, snapshot)
...
+      def self.detect_bump_type(version_name_a, version_name_b)
+        if version_name_a == version_name_b
+          return :none
+        end
+
+        version_a = Gem::Version.new(version_name_a)
+        version_b = Gem::Version.new(version_name_b)
+
+        same_major = version_a.canonical_segments[0] == version_b.canonical_segments[0]
+        same_minor = version_a.canonical_segments[1] == version_b.canonical_segments[1]
```
tonidero, line 119: This would work as long as both versions have 3 canonical segments. Maybe we should validate that? NABD though, since this is internal.

lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_action.rb
```diff
@@ -106,6 +95,23 @@ def self.available_options
...
         true
       end
+
+      private_class_method def self.open_pr_against_main(automatic_release, github_pr_token, new_branch_name, new_version_number, repo_name, version_number)
+        Helper::RevenuecatInternalHelper.commit_changes_and_push_current_branch("Version bump for #{new_version_number}")
+
+        pr_title = "Updates purchases-hybrid-common to #{new_version_number}"
+        type_of_bump = Helper::VersioningHelper.detect_bump_type(version_number, new_version_number)
+        labels = ['dependencies']
+        labels << 'minor' if type_of_bump == :minor
```
tonidero, line 105: Hmm I was thinking about adding the `major` label for majors, but then again, we are not autobumping for majors I believe so not needed 👍

tonidero (approved):

PR #42, opened 2022-11-28
Include links to native dependencies

vegaro:
[CSDK-514]

[CSDK-514]: https://revenuecats.atlassian.net/browse/CSDK-514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spec/helper/versioning_helper_spec.rb
```diff
@@ -29,6 +58,17 @@
     let(:get_commit_no_items) do
       { body: File.read("#{File.dirname(__FILE__)}/../test_files/get_commit_sha_4ceaceb20e700b92197daf8904f5c4e226625d8a.json") }
     end
+    let(:get_commits_response_hybrid) do
+      { body: File.read("#{File.dirname(__FILE__)}/../test_files/get_commits_since_last_release_hybrid.json") }
```
vegaro, line 62: This is a real scenario response obtained when running `purchases-flutter`'s bump lane, I thought it would be better to test using real cases, that's why all the new responses

lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
```diff
@@ -40,6 +40,39 @@ def self.get_commits_since_old_version(github_token, old_version, repo_name)
...
+        all_releases = JSON.parse(response[:body])
+
+        # Filters releases between tags
+        all_releases.select do |release|
+          version = Gem::Version.new(release["tag_name"])
+          start_tag < version && version <= end_tag
+        end
+      end
+
+      private_class_method def self.get_github_releases(repo_name) end
```
tonidero, line 75: Hmm this doesn't seem to be used?

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -2,6 +2,8 @@
 require 'fastlane/action'
 require 'fastlane/actions/github_api'
 require_relative 'github_helper'
+require_relative '../actions/bump_phc_version_action'
```
tonidero, line 5: hmm I'm not sure we should import an action from a helper... I know we are just using a constant there. Should we just move it to another file that can be accessed from both? Otherwise, I prefer moving the constant to the helper and have the action access it instead.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -21,6 +23,11 @@ module Fastlane
 
   module Helper
     class VersioningHelper
+      LATEST_VERSION_COLUMN = 1
+      IOS_VERSION_COLUMN = 2
+      ANDROID_VERSION_COLUMN = 3
+      PHC_VERSION_COLUMN = 4
```
tonidero, line 29: Note that some hybrids (Unity) can have additional columns

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -61,6 +68,10 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
             section = get_section_depending_on_types_of_change(types_of_change)
 
             line = "* #{message} via #{name} (@#{username})"
+            if message.include?(Fastlane::Actions::BumpPhcVersionAction::PR_TITLE)
```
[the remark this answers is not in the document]
tonidero: Hmm I wonder if creating a tag would be safer... but it's not that big of a difference so I'm ok with this for now.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -173,6 +194,8 @@ def self.detect_bump_type(version_name_a, version_name_b)
           :new_features
         elsif change_types.include?("fix")
           :fixes
+        elsif change_types.include?("dependencies")
```
tonidero, line 197: I think we need to be careful with how we use the `dependencies` label after this... How about renaming it to `phc_dependencies` or something like that to make it more explicit? Otherwise I foresee us using this tag for some unrelated dependency.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -76,6 +87,14 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
         build_changelog_sections(changelog_sections)
       end
 
+      def self.platform_changelogs(releases, platform)
+        platform_changelogs = []
+        releases.each do |release|
+          platform_changelogs.push("  * (#{platform} #{release['name']})[#{release['html_url']}]")
```
tonidero, line 93: I wonder if we should create a `Release` data holder to contain the data we need instead of passing around the parsed json... I'm ok with this for now though since we follow the same pattern in a lot of places, and we probably don't need that yet.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -225,6 +248,25 @@ def self.get_type_of_bump_from_commits(commits, github_token, rate_limit_sleep,
...
+        previous_ios_version = versions_latest_release[IOS_VERSION_COLUMN]
+        previous_android_version = versions_latest_release[ANDROID_VERSION_COLUMN]
+
+        new_android_version = Helper::UpdateHybridsVersionsFileHelper.get_android_version_for_hybrid_common_version(phc_version)
+        UI.message("Obtained android version #{new_android_version} for PHC version #{phc_version}")
+        new_ios_version = Helper::UpdateHybridsVersionsFileHelper.get_ios_version_for_hybrid_common_version(phc_version)
+        UI.message("Obtained ios version #{new_ios_version} for PHC version #{phc_version}")
+
+        android_releases = Helper::GitHubHelper.get_releases_between_tags(github_token, previous_android_version, new_android_version, 'purchases-android')
+        ios_releases = Helper::GitHubHelper.get_releases_between_tags(github_token, previous_ios_version, new_ios_version, 'purchases-ios')
```
tonidero, line 263: Should we have a `constants.rb` file or something like that to hold these repo names?

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -61,6 +68,10 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep)
             section = get_section_depending_on_types_of_change(types_of_change)
 
             line = "* #{message} via #{name} (@#{username})"
+            if message.include?(Fastlane::Actions::BumpPhcVersionAction::PR_TITLE)
+              # Append links to native releases
+              line += native_releases_links(github_token, hybrid_common_version, versions_file_path)
```
tonidero, line 73: So if we bump PHC multiple times for the same version (shouldn't happen often) this will add the same entries multiple times right? 

Thinking of possible solutions, I guess we should do this not in a per-commit basis but just getting the last phc bump since the last release and do the comparison. It might be tricky though. What do you think?

tonidero (commented):
Haven't reviewed tests yet but I thought I would post what I have. Really excited about this!

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: I've actually encountered that in the past and I manually removed the extra changelog line to prevent something like:

```
* [AUTOMATIC BUMP] Updates purchases-hybrid-common to 4.5.3 (#553) via RevenueCat Git Bot (@RCGitBot)
* [AUTOMATIC BUMP] Updates purchases-hybrid-common to 4.5.2 (#553) via RevenueCat Git Bot (@RCGitBot)
```

This might be a good opportunity to fix it. I wasn't sure how to do it but if we start using `phc_dependencies`, we might be able to do it easily

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
joshdholtz: This isn't an end of the world thing. I'd be fine if this is how it stayed 😇  But... might be nice to have a constants file for if/when other constants like this are added 🤷‍♂️

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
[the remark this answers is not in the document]
joshdholtz: This might be a nice comment to add into the code as well ☝️

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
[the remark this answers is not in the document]
joshdholtz: I would maybe change this to `start_with?` but looks fine!

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
joshdholtz: I agree with a `Release` data holder idea at some point! ☝️

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
joshdholtz: Yes yes, I agree with that idea 😊 This might be a good place to also put that `PR_TITLE` constant as well

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -225,6 +248,25 @@ def self.get_type_of_bump_from_commits(commits, github_token, rate_limit_sleep,
...
+        UI.message("Obtained android version #{new_android_version} for PHC version #{phc_version}")
+        new_ios_version = Helper::UpdateHybridsVersionsFileHelper.get_ios_version_for_hybrid_common_version(phc_version)
+        UI.message("Obtained ios version #{new_ios_version} for PHC version #{phc_version}")
+
+        android_releases = Helper::GitHubHelper.get_releases_between_tags(github_token, previous_android_version, new_android_version, 'purchases-android')
+        ios_releases = Helper::GitHubHelper.get_releases_between_tags(github_token, previous_ios_version, new_ios_version, 'purchases-ios')
+
+        native_dependency_changelogs = [""]
+        native_dependency_changelogs.concat(platform_changelogs(android_releases, 'Android'))
+        native_dependency_changelogs.concat(platform_changelogs(ios_releases, 'iOS'))
```
joshdholtz, line 267: More Ruby-ish way 👇 

```suggestion
        native_dependency_changelogs += platform_changelogs(android_releases, 'Android')
        native_dependency_changelogs += platform_changelogs(ios_releases, 'iOS')
```

spec/test_files/VERSIONS.md
```diff
@@ -0,0 +1,8 @@
+| Version | iOS version | Android version | Common files version |
+|---------|-------------|-----------------|----------------------|
+| 4.5.3   | 4.15.2      | 5.6.5           | 4.5.2                |
+| 4.5.2   | 4.15.2      | 5.6.5           | 4.5.1                |
```
joshdholtz, line 4: I think @tonidero mentioned something about Unity having a different number of columns so might be nice to have a Unity style table in the test (and maybe one that is just super messed up to test handling of errors gracefully) 🤷‍♂️

joshdholtz (changes requested):
This looks good! Left some comments and feedback but otherwise looks 💯

spec/helper/versioning_helper_spec.rb
```diff
@@ -1,3 +1,20 @@
+def mock_native_releases
```
tonidero, line 1: Hmm I think I would move this inside the `describe`, since it seems it's only used there. Probably at the end.

spec/test_files/purchases_android_releases.json
```diff
@@ -0,0 +1,2074 @@
+[
```
tonidero, line 1: I wonder, would it make sense to add a comment to the top of this JSON specifiying the exact request done to get these (without any secrets if any) and date? I think these would be very helpful to track how these were obtained. Alternatively we could rename the file with more details, but I think the comment might be easier.

spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
```diff
@@ -167,7 +167,7 @@ def setup_stubs
       allow(Fastlane::Helper::RevenuecatInternalHelper).to receive(:validate_local_config_status_for_bump)
         .with('release/1.13.0', mock_github_pr_token)
       allow(Fastlane::Helper::VersioningHelper).to receive(:auto_generate_changelog)
-        .with(mock_repo_name, mock_github_token, 3)
+        .with(mock_repo_name, mock_github_token, 3, nil, nil)
```
tonidero, line 170: We should add a test making sure the parameters are passed correctly if they are passed.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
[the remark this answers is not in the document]
vegaro: changed this to use the label

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: I think we can do this as a followup if you don't mind

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: changed it with a new label

spec/test_files/VERSIONS.md
vegaro: Added some extra tests

spec/test_files/purchases_android_releases.json
vegaro: Added a comment in this one and purchases_android_releases

lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
```diff
@@ -8,7 +8,7 @@ module Fastlane
 
   module Helper
     class GitHubHelper
-      SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf refactor style test next_release dependencies minor].to_set
+      SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf refactor style test next_release dependencies phc_dependencies minor].to_set
```
tonidero, line 11: Nothing to do in this PR but we will need to add this label to the list of valid labels in danger

tonidero (approved):
Looks great!

PR #43, opened 2022-12-29
Use last commit instead of HEAD when comparing against last release

vegaro:
We should be comparing against the commit where the lane is running and not against HEAD. It's possible HEAD is actually not the commit where the lane is running at.

### DIFF
--- a/spec/helper/github_helper_spec.rb
+++ b/spec/helper/github_helper_spec.rb
@@ -64,9 +64,13 @@ describe Fastlane::Helper::GitHubHelper do
     end
 
     it 'returns commits from response' do
+      allow(Fastlane::Actions::LastGitCommitAction).to receive(:run)
+        .and_return({
+                                                                       commit_hash: 'cfdd80f73d8c91121313d72227b4cbe283b57c1e'
+                                                                     })
       allow(Fastlane::Actions::GithubApiAction).to receive(:run)
         .with(server_url: server_url,
-              path: '/repos/RevenueCat/mock-repo-name/compare/1.11.0...HEAD',
+              path: '/repos/RevenueCat/mock-repo-name/compare/1.11.0...cfdd80f73d8c91121313d72227b4cbe283b57c1e',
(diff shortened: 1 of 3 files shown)


spec/helper/versioning_helper_spec.rb
```diff
@@ -517,4 +477,18 @@ def setup_commit_search_stubs(hashes_to_responses)
         .and_return(response)
     end
   end
+
+  def mock_commits_since_last_release(last_commit_hash, response)
+    allow(Fastlane::Actions::LastGitCommitAction).to receive(:run)
+      .and_return({
+                                                                     commit_hash: last_commit_hash
+                                                                   })
```
aboedo, line 485: indentation seems off here. Also, do we need the `()` for this one? not sure if we have a standard for this already but if feels like this would look potentially nicer:
```suggestion
        allow(Fastlane::Actions::LastGitCommitAction).to receive(:run).and_return(commit_hash: last_commit_hash)
```

spec/helper/github_helper_spec.rb
```diff
@@ -64,9 +64,13 @@
     end
 
     it 'returns commits from response' do
+      allow(Fastlane::Actions::LastGitCommitAction).to receive(:run)
+        .and_return({
+                                                                       commit_hash: 'cfdd80f73d8c91121313d72227b4cbe283b57c1e'
```
aboedo, line 69: should probably extract this string

aboedo, line 69: I left a suggestion in the next file for improving the readability for this

aboedo (approved):
looks good! left a couple of suggestions

tonidero (approved):
After Andy's suggestions :shipit:

spec/helper/versioning_helper_spec.rb
vegaro: `rubocop` formatted it like this, but I like the suggestion, didn't know that was possible! I will update

PR #45, opened 2023-01-18
`Changelog`: added new "Performance Improvements" section

NachoSoto:
For example, https://github.com/RevenueCat/purchases-ios/releases/tag/4.17.0 had a bunch of nice performance improvements that would have been nice to call out.
Future releases will have that now.


aboedo (approved):
awesome! We need to add a label and add it to danger too right?

NachoSoto:
We already had it. The only change is that it used to be grouped as part of "other" but now they're in their own section.

NachoSoto:
For example: https://github.com/RevenueCat/purchases-ios/pull/2199

NachoSoto:
What's the process for updating this in all repos? 🙏🏻

aboedo:
I believe we need a PR _per repo_ that looks like this: https://github.com/RevenueCat/purchases-ios/pull/2194/files

tonidero:
Right, unfortunately right now you need to run a [bundle update](https://www.notion.so/revenuecat/CI-Integration-Tests-Automation-db61f2b3f1fc469bbd6f23c04fa1a3a1#c04f8a29ba13414b954d9266333aaf1a) on each repo

We have plans to actually release the plugin to rubygems so we can do version specifiers and avoid this. There is a task for that

NachoSoto:
Running `bundle update` leads to an infinite loop printing "Resolving dependencies..." lol

tonidero:
Ahhh I think it's `bundle update fastlane-plugin-revenuecat_internal`

NachoSoto:
That worked thanks!

PR #47, opened 2023-01-31
Improve prerelease changelog generation

joshdholtz:
⚠️ This still needs new tests for when prerelease is true (currently is just passing with old tests)

## Motivation

Improve `fastlane bump` for prereleases

This is fixing an issue when bumping from `6.0.0-alpha.1` to `6.0.0-alpha.2`
- The generated changelog was empty which was 💥 ing the entire process

## Description

Can now run `fastlane bump is_prerelease:true`

See https://github.com/RevenueCat/purchases-android/pull/764 for results

- New `is_prerelease` option on `bump_version_update_changelog_create_pr`
    - Defaults to `false` 
- Renamed `create_pr_to_main` to `create_pr`
    - Added `base_branch` as a parameter
- Renamed `latest_non_prerelease_version_number` to `latest_version_number`
    - Added `include_prerelease` parameter
    - This is what allow for a nice changelog only between prerelease versions
- `get_pr_resp_items_for_sha` no longer default to `main` as the base branch

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_action.rb
@@ -110,7 +110,7 @@ module Fastlane
           pr_title = "[AUTOMATIC BUMP] #{pr_title}"
         end
 
-        Helper::RevenuecatInternalHelper.create_pr_to_main(pr_title, body, repo_name, new_branch_name, github_pr_token, labels)
+        Helper::RevenuecatInternalHelper.create_pr(pr_title, body, repo_name, "main", new_branch_name, github_pr_token, labels)
       end
     end
   end
--- a/lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_action.rb
@@ -104,13 +104,14 @@ module Fastlane
         labels = ['phc_dependencies']
         labels << 'minor' if type_of_bump == :minor
         body = pr_title
+        base_branch = "main"
 
         if automatic_release
           body = "**This is an automatic bump.**\n\n#{body}"
           pr_title = "[AUTOMATIC BUMP] #{pr_title}"
         end
 
-        Helper::RevenuecatInternalHelper.create_pr(pr_title, body, repo_name, "main", new_branch_name, github_pr_token, labels)
+        Helper::RevenuecatInternalHelper.create_pr(pr_title, body, repo_name, base_branch, new_branch_name, github_pr_token, labels)
       end
     end
   end
--- a/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
@@ -21,7 +21,7 @@ module Fastlane
         automatic_release = params[:automatic_release]
         hybrid_common_version = params[:hybrid_common_version]
         versions_file_path = params[:versions_file_path]
-        include_prerelease = params[:is_prerelease]
+        include_prereleases = params[:is_prerelease]
 
         current_branch = Actions.git_branch
         if UI.interactive? && !UI.confirm("Current branch is #{current_branch}. Are you sure this is the base branch for your bump?")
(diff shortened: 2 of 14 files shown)


lib/fastlane/plugin/revenuecat_internal/actions/bump_phc_version_action.rb
```diff
@@ -110,7 +110,7 @@ def self.is_supported?(platform)
           pr_title = "[AUTOMATIC BUMP] #{pr_title}"
         end
 
-        Helper::RevenuecatInternalHelper.create_pr_to_main(pr_title, body, repo_name, new_branch_name, github_pr_token, labels)
+        Helper::RevenuecatInternalHelper.create_pr(pr_title, body, repo_name, "main", new_branch_name, github_pr_token, labels)
```
tonidero, line 113: To avoid the hardcoded `main`, maybe we can extract to a constant?

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -21,6 +21,7 @@ def self.run(params)
         automatic_release = params[:automatic_release]
         hybrid_common_version = params[:hybrid_common_version]
         versions_file_path = params[:versions_file_path]
+        include_prerelease = params[:is_prerelease]
```
tonidero, line 24: Just to check, this will be passed from CircleCI as a parameter that needs to be added manually, correct? After this is done, we probably should update notion to include these new steps for prereleases

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -160,13 +160,18 @@ def self.detect_bump_type(version_name_a, version_name_b)
...
       end
 
-      private_class_method def self.latest_non_prerelease_version_number
-        Actions
-          .sh("git tag", log: false)
-          .strip
-          .split("\n")
-          .select { |tag| tag.match("^[0-9]+.[0-9]+.[0-9]+$") }
-          .max_by { |tag| Gem::Version.new(tag) }
+      private_class_method def self.latest_version_number(include_prerelease: false)
```
tonidero, line 163: I've been feeling as I read this that I like the plural `include_prereleases` more. Thoughts?

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -145,7 +146,12 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :versions_file_path,
                                        description: "Path to the VERSIONS.md file",
                                        optional: true,
-                                       type: String)
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :is_prerelease,
```
tonidero, line 150: The parameter everywhere is called `include_prerelease`. Should we rename the parameter to match?

lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
```diff
@@ -16,9 +16,11 @@ def self.get_pr_resp_items_for_sha(sha, github_token, rate_limit_sleep, repo_nam
           sleep(rate_limit_sleep)
         end
 
+        current_branch = Actions.git_branch
```
tonidero, line 19: Hmm this trusts that we are in the correct branch before this. I wonder if we should be passing the base branch as a parameter instead to make it more explicit... This probably works for now so we can change it in the future if we have to.

tonidero (commented):
Looking good, just some small things for now

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
joshdholtz: I... completely agree with that! Will change 😊

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
joshdholtz: I was torn because this isn't about including prereleases but saying this bump _is_ a prerelease 🤷‍♂️ So I _think_ this one is okay to stay different? 🤔 Thoughts?

lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
joshdholtz: OHHHHH, I totally meant to make that change but did this to get things to work 🙈 Thanks for catching!

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
joshdholtz: @tonidero Ah, so no! Nothing should need to get updated from CircleCI since this is really only meant to be run locally for one-off prerelease builds. `is_prerelease` defaults to `false` so everything should work as it does right now 💪

tonidero: We could potentially also run this from CircleCI but manually passing a parameter to the job, but that would probably be a bit more work and might not be worth it. I think we can just document how to make prerelease releases locally in notion

tonidero (approved):
LGTM!

PR #46, opened 2023-01-25
Add pattern option to files to update

vegaro:
[CSDK-644](https://revenuecats.atlassian.net/browse/CSDK-644)

This will prevent updating the wrong versions when replacing in files

[CSDK-644]: https://revenuecats.atlassian.net/browse/CSDK-644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

The way it should be called now would be passing the following to the action:

```
files_with_version_number = {
  './.version' => ["{x}"],
  './pubspec.yaml' => ["version: {x}"],
  './ios/purchases_flutter.podspec' => ["s.version          = '{x}'"],
  './macos/purchases_flutter.podspec' => ["s.version          = '{x}'", "s.dependency_version     = '{x}'"],
  './ios/Classes/PurchasesFlutterPlugin.m' => ["return @\"{x}\""],
  './android/build.gradle' => ["version '{x}'"],
  './android/src/main/java/com/revenuecat/purchases_flutter/PurchasesFlutterPlugin.java' => ["PLUGIN_VERSION = \"{x}\""],
}
 ```
 We should update it in all actions that call `replace_version_number`

PR #50, opened 2023-04-26
Fix replace_version_number_action to accept a hash instead of an array of files

vegaro:
With the changes from #46 , the helper `replace_version_number` got updated to accept a hash of files instead of an array. `replace_version_number_action` depends on this helper, so the action needed to be updated. This PR updates the parameters and fixes some tests

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/replace_version_number_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/replace_version_number_action.rb
@@ -36,15 +36,22 @@ module Fastlane
                                        type: String),
           FastlaneCore::ConfigItem.new(key: :files_to_update,
                                        env_name: "RC_INTERNAL_FILES_TO_UPDATE_VERSION",
-                                       description: "Files that contain the version number and need to have it updated",
+                                       description: 'Hash of files that contain the version number and need to have it' \
+                                                    'updated to the patterns that contains the version in the file.' \
+                                                    'Mark the version in the pattern using {x}.' \
+                                                    'For example: { "./pubspec.yaml" => ["version: {x}"] }',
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/revenuecat_internal/actions/replace_version_number_action.rb
```diff
@@ -36,15 +36,22 @@ def self.available_options
...
           FastlaneCore::ConfigItem.new(key: :files_to_update,
                                        env_name: "RC_INTERNAL_FILES_TO_UPDATE_VERSION",
-                                       description: "Files that contain the version number and need to have it updated",
+                                       description: 'Hash of files that contain the version number and need to have it' \
+                                                    'updated to the patterns that contains the version in the file.' \
+                                                    'Mark the version in the pattern using {x}.' \
+                                                    'For example: { "./pubspec.yaml" => ["version: {x}"] }',
                                        optional: false,
-                                       type: Array),
+                                       type: Hash),
```
tonidero, line 44: I was trying to look in the documentation, is this an accepted type? I couldn't find it in the documentation... The closes I could find was polymorphic arguments: https://docs.fastlane.tools/advanced/actions/

But well, if you've tested it and it works, it should be ok.

tonidero (approved):

lib/fastlane/plugin/revenuecat_internal/actions/replace_version_number_action.rb
vegaro: I couldn't find it either! But I saw it being used in their repo https://github.com/search?q=repo:fastlane/fastlane%20type:%20Hash&type=code So I assumed they just hadn't documented it

PR #51, opened 2023-05-19
Don't ensure Git status clean when running on CI

vegaro:
When running on CI we should reset all files instead of failing the lane if there are changes.

It's possible the Git repository gets dirty after running `npm install` or `bundle install` and the lock files get updated.


NachoSoto (approved):
♥️

tonidero (approved):

PR #54, opened 2023-08-24
Add support for `RevenueCatUI` label

NachoSoto:
This will allow PRs with this label to be grouped together instead of all inside "Other changes".
If a PR has `feat` at the same time, that gets prioritized, but a PR with something like `refactor` + `RevenueCatUI` will use this new logic.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
@@ -8,7 +8,7 @@ module Fastlane
 
   module Helper
     class GitHubHelper
-      SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf RevenueCatUI refactor style test next_release dependencies phc_dependencies minor].to_set
+      SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf revenuecatui refactor style test next_release dependencies phc_dependencies minor].to_set
 
       def self.get_pr_resp_items_for_sha(sha, github_token, rate_limit_sleep, repo_name, base_branch)
         if rate_limit_sleep > 0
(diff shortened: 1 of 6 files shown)


spec/helper/versioning_helper_spec.rb
```diff
@@ -309,6 +315,8 @@
       )
       expect(changelog).to eq("### New Features\n" \
                               "* added a log when `autoSyncPurchases` is disabled (#1749) via aboedo (@aboedo)\n" \
+                              "### Paywalls\n" \
```
NachoSoto, line 318: Should we name this `RevenueCatUI` instead?

lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
```diff
@@ -8,7 +8,7 @@ module Fastlane
 
   module Helper
     class GitHubHelper
-      SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf refactor style test next_release dependencies phc_dependencies minor].to_set
+      SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf RevenueCatUI refactor style test next_release dependencies phc_dependencies minor].to_set
```
tonidero, line 11: Hmm should we make it lowercase? Like `revenuecatui`. Since all the others are lowercase

spec/helper/versioning_helper_spec.rb
tonidero: Hmm I would match it with the name of the library, so there is less confusion. So I would say yes?

tonidero (approved):
LGTM!

lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
NachoSoto: Done, and I renamed the label on the iOS repo.

PR #58, opened 2024-01-29
Stop adding `minor` labels if minor phc update

vegaro:
We don't want to make the type of update dependent of phc anymore


tonidero (approved):

tonidero:
Ah looks like we need to update the tests

PR #59, opened 2024-03-13
Add `trigger_action_in_circle_ci`

vegaro:
Useful to run bumps from fastlane from the repositories

```
desc "Trigger bump"
lane :trigger_bump do
    trigger_action_in_circle_ci(action: 'bump', repo_name: repo_name)
end
  ```

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/trigger_action_in_circle_ci_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/trigger_action_in_circle_ci_action.rb
@@ -6,15 +6,13 @@ require 'json'
 
 module Fastlane
   module Actions
-    class TriggerActionInCircleCI < Action
+    class TriggerActionInCircleCiAction < Action
       def self.run(params)
         circle_token = params[:circle_token]
         UI.user_error!("Please set the CIRCLE_TOKEN environment variable") unless circle_token
 
-        branch = params[:branch]
         default_branch = Actions.git_branch
-        branch = UI.input("Branch (defaults to #{default_branch}): ")
-        branch = default_branch if branch.empty?
+        branch = params[:branch] || default_branch
 
         headers = {"Circle-Token" => circle_token, "Content-Type" => "application/json", "Accept" => "application/json"}
         data = {parameters: {'action' => params[:action]}, branch: branch}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/fastlane/plugin/revenuecat_internal/actions/trigger_action_in_circle_ci_action.rb [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+module Fastlane
+  UI = FastlaneCore::UI unless Fastlane.const_defined?(:UI)
+
+  module Actions
+    class TriggerActionInCircleCiAction < Action
+      def self.run(params)
+        circle_token = params[:circle_token]
+        UI.user_error!("Please set the CIRCLE_TOKEN environment variable") unless circle_token
+
+        default_branch = Actions.git_branch
```
tonidero, line 16: Rather than the `default_branch` this would be the `current_branch` right? As in, it will use whatever branch it's currently in, I think.

lib/fastlane/plugin/revenuecat_internal/actions/trigger_action_in_circle_ci_action.rb
```diff
@@ -0,0 +1,71 @@
...
+require 'json'
+
+module Fastlane
+  UI = FastlaneCore::UI unless Fastlane.const_defined?(:UI)
+
+  module Actions
+    class TriggerActionInCircleCiAction < Action
+      def self.run(params)
+        circle_token = params[:circle_token]
+        UI.user_error!("Please set the CIRCLE_TOKEN environment variable") unless circle_token
```
tonidero, line 14: Should we also verify that the action is passed as a parameter?

tonidero (approved):
Couple comments but not blocking. LGTM!

joshdholtz (approved):

lib/fastlane/plugin/revenuecat_internal/actions/trigger_action_in_circle_ci_action.rb
```diff
@@ -10,14 +10,18 @@ module Fastlane
   module Actions
     class TriggerActionInCircleCiAction < Action
       def self.run(params)
+        action = params[:action]
```
tonidero, line 13: I was just thinking that it could be nice to accept random parameters for this... So for example we can run jobs that require passing a version number or things like that... But this is ok for now!

vegaro: Oh so we could add parameters to the CircleCI jobs that will eventually get passed to Fastlane?  I like it

PR #60, opened 2024-03-14
Ask for branch when triggering action

vegaro:
Followup on #59 asking for branch and displaying error if error in request


lib/fastlane/plugin/revenuecat_internal/actions/trigger_action_in_circle_ci_action.rb
```diff
@@ -17,15 +17,20 @@ def self.run(params)
         circle_token = params[:circle_token]
         UI.user_error!("Please set the CIRCLE_TOKEN environment variable") unless circle_token
 
+        branch = params[:branch]
         current_branch = Actions.git_branch
-        branch = params[:branch] || current_branch
+        if branch.nil? || branch.empty?
+            branch = UI.input("Branch (defaults to #{current_branch}): ")
+        end
+        branch = current_branch if branch == ""
```
tonidero, line 25: Hmm I don't remember but if we make the parameter not optional, I think fastlane would automatically ask for the branch if not passed? I'm not sure if it would allow an empty value though, so this might be better?

tonidero (approved):

lib/fastlane/plugin/revenuecat_internal/actions/trigger_action_in_circle_ci_action.rb
vegaro: you're correct! I will clean it up

PR #61, opened 2024-04-17
Support updating files on latest stable release only

tonidero:
When we need to make a hotfix or make a prerelease, sometimes we update the files that we shouldn't. This adds support for a new parameter on these lanes that will only be updated if the version given is posterior (in semver versions) to the latest tag and it's not a prerelease.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -27,14 +27,14 @@ module Fastlane
         files_to_update_without_prerelease_modifiers.each do |file_to_update, patterns|
           replace_in(previous_version_number_without_prerelease_modifiers, new_version_number_without_prerelease_modifiers, file_to_update, patterns)
         end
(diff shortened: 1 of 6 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -23,6 +27,19 @@ def self.replace_version_number(previous_version_number, new_version_number, fil
...
           replace_in(previous_version_number_without_prerelease_modifiers, new_version_number_without_prerelease_modifiers, file_to_update, patterns)
         end
+        if !files_to_update_on_latest_stable_releases.empty? && latest_semver_tag?(new_version_number) && !Gem::Version.new(new_version_number).prerelease?
+          files_to_update_on_latest_stable_releases.each do |file_to_update, patterns|
+            replace_in(previous_version_number_without_prerelease_modifiers, new_version_number_without_prerelease_modifiers, file_to_update, patterns)
+          end
+        end
+      end
+
+      def self.latest_semver_tag?(version_number)
```
vegaro, line 37: I would rename this to "newer_than_latest_published_version" since it doesn't require `version_number` to be a tag

PR #63, opened 2024-04-24
Fix issue where we were not detecting versions as latest if there was a higher prerelease version

tonidero:
This was causing problems when we had some prerelease for a future version shipped. In that case, we don't want to consider that one the latest.

### DIFF
--- a/spec/helper/revenuecat_internal_helper_spec.rb
+++ b/spec/helper/revenuecat_internal_helper_spec.rb
@@ -10,7 +10,7 @@ describe Fastlane::Helper::RevenuecatInternalHelper do
 
     before(:each) do
       allow(Fastlane::Actions).to receive(:sh).with("git fetch --tags -f")
-      allow(Fastlane::Actions).to receive(:sh).with("git tag -l '[0-9]*.[0-9]*.[0-9]*' | sort -r --version-sort | head -n1").and_return('1.2.3')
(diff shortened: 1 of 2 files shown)


spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -217,11 +217,11 @@
   describe '.newer_than_latest_published_version?' do
     before(:each) do
       allow(Fastlane::Actions).to receive(:sh).with("git fetch --tags -f")
-      allow(Fastlane::Actions).to receive(:sh).with("git tag -l '[0-9]*.[0-9]*.[0-9]*' | sort -r --version-sort | head -n1").and_return('1.2.3')
+      allow(Fastlane::Actions).to receive(:sh).with("git tag | grep '^[0-9]*\.[0-9]*\.[0-9]*$' | sort -r --version-sort | head -n1").and_return('1.2.3')
     end
 
     it 'if no tag is returned its considered latest' do
-      allow(Fastlane::Actions).to receive(:sh).with("git tag -l '[0-9]*.[0-9]*.[0-9]*' | sort -r --version-sort | head -n1").and_return('')
+      allow(Fastlane::Actions).to receive(:sh).with("git tag | grep '^[0-9]*\.[0-9]*\.[0-9]*$' | sort -r --version-sort | head -n1").and_return('')
```
vegaro, line 224: mind extracting the string now that we are at it?

vegaro (approved):
Good catch and fix!

PR #64, opened 2024-05-06
ReplaceVersionNumber: Make stable version substitution work with versions not matching previous version

tonidero:
We noticed that the files that we were using the latest stable versions weren't being updated correctly because the "previous_version" did not match the previous version that we give to the job, since that would be the version of the latest stable version at that point in time. 

This allows that piece of substitution to work with any stable semver version instead of having to specific the previous version specifically.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
@@ -116,7 +116,8 @@ module Fastlane
                                        env_name: "RC_INTERNAL_FILES_TO_UPDATE_ON_LATEST_STABLE_RELEASES",
                                        description: 'Hash of files that contain the version number and only need to' \
                                                     'be updated on stable releases (no prereleases) and on the latest' \
-                                                    'major (no hotfixes).' \
+                                                    'major (no hotfixes). Note that the version will be updated as' \
+                                                    'long as it matches a semver stable version, even if its not the previous version' \
                                                     'Mark the version in the pattern using {x}.' \
                                                     'For example: { "./pubspec.yaml" => ["version: {x}"] }',
                                        optional: true,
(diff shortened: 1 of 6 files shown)


tonidero:
Note that we will also need to update the files in the repos using this to make sure they have stable versions, and not snapshots or anything like that, in order for this to work.

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -116,7 +116,8 @@ def self.available_options
                                        env_name: "RC_INTERNAL_FILES_TO_UPDATE_ON_LATEST_STABLE_RELEASES",
                                        description: 'Hash of files that contain the version number and only need to' \
                                                     'be updated on stable releases (no prereleases) and on the latest' \
-                                                    'major (no hotfixes).' \
+                                                    'major (no hotfixes). Note that the version will be updated as' \
+                                                    'long as it matches a semver stable version, even if its not the previous version' \
```
aboedo, line 120: should all of these have spaces at the end of every sentence? I don't think ruby would auto-add spaces. Feels like we've had this for all actions

spec/helper/revenuecat_internal_helper_spec.rb
```diff
@@ -214,6 +214,38 @@
...
+        '1.2.3',
+        '1.2.4',
+        {},
+        {},
+        {
+          file_to_update_on_latest_stable_release_5 => ["url=https://sdk.revenuecat.com/android/{x}/index.html"]
+        }
+      )
+
+      expect(File.read(file_to_update_on_latest_stable_release_5)).to eq('<meta http-equiv="refresh" content="0; url=https://sdk.revenuecat.com/android/1.2.4/index.html" />')
```
aboedo, line 231: not super against this, but it feels like we could make this a bit easier to parse by either string formatting the version or breaking the expected and actual strings into their own variables and then comparing them? 
like
```suggestion
      original = '<meta http-equiv="refresh" content="0; url=https://sdk.revenuecat.com/android/7.10.7/index.html" />'
      expected = '<meta http-equiv="refresh" content="0; url=https://sdk.revenuecat.com/android/1.2.4/index.html" />'
      actual = File.read(file_to_update_on_latest_stable_release_5)
      expect(actual).to eq(expected)
```

(or use string formatting to make the version a variable)

aboedo (approved):

vegaro (approved):

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
tonidero: Good catch! Yeah that's right 👍

PR #65, opened 2024-05-30
Sync to spm repo

jamesrb1:
Use fastlane to sync the iOS repo to its SPM variant. This is used by the changes made in this [purchases-ios PR](https://github.com/RevenueCat/purchases-ios/pull/3926).

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
@@ -1,0 +1,39 @@
+require 'fastlane/action'
+require 'fastlane_core/configuration/config_item'
+require_relative '../helper/revenuecat_internal_helper'
+require_relative '../helper/versioning_helper'
+
+module Fastlane
+  module Actions
+    class GitCloneAndPushAction < Action
+      def self.run(params)
+        source_repo = params[:source_repo]
+        destination_repo = params[:destination_repo]
+
+        Helper::RevenuecatInternalHelper.git_clone_source_to_dest(source_repo, destination_repo)
+      end
+
+      def self.description
+        "Makes a lightweight clone containing only on the main branch's history."
+      end
+
+      def self.authors
+        ["James Borthwick"]
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :source_repo,
+                                       description: "The source repository URL",
+                                       is_string: true),
+          FastlaneCore::ConfigItem.new(key: :destination_repo,
+                                       description: "The destination repository URL",
+                                       is_string: true)
+        ]
+      end
+
+      def self.is_supported?(platform)
+        [:ios, :mac].include?(platform)
+      end
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
```diff
@@ -0,0 +1,40 @@
...
+                                       description: "The source repository URL",
+                                       is_string: true),
+          FastlaneCore::ConfigItem.new(key: :destination_repo,
+                                       description: "The destination repository URL",
+                                       is_string: true)
+        ]
+      end
+
+      def self.is_supported?(platform)
+        [:ios, :mac].include?(platform)
```
tonidero, line 36: I think we can just hardcode this to `true`? 
```suggestion
        true
```

lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      def self.authors
+        ["James Borthwick"]
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :source_repo,
+                                       description: "The source repository URL",
+                                       is_string: true),
```
tonidero, line 28: I don't remember if `type: String` is also required (I see it in the other actions). But if it works, I think it should be ok.

lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
```diff
@@ -0,0 +1,40 @@
...
+    class GitCloneAndPushAction < Action
+      def self.run(params)
+        source_repo = params[:source_repo]
+        destination_repo = params[:destination_repo]
+
+        Helper::RevenuecatInternalHelper.git_clone_source_to_dest(source_repo, destination_repo)
+      end
+
+      def self.description
+        "Makes a lightweight clone containing only on the main branch's history."
```
tonidero, line 17: Hmm it also contains the tags right?

lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
```diff
@@ -0,0 +1,40 @@
+require 'fastlane/action'
+require 'fastlane_core/configuration/config_item'
+require_relative '../helper/revenuecat_internal_helper'
+require_relative '../helper/versioning_helper'
+
+module Fastlane
+  module Actions
+    class GitCloneAndPushAction < Action
```
tonidero, line 8: Might be good to add some tests, even if there is basically no logic. Mostly to make sure the parameters and such don't change without us knowing.

tonidero (approved):
Some non-blocker comments. I think the main one is about adding tests, but can happen in a different PR.

lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
jamesrb1: Apparently `is_string` is an outdated way of doing this. I've replaced it with `type: String`, and added `optional: false` as well.

lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
jamesrb1: Yes, I will add that info too.

lib/fastlane/plugin/revenuecat_internal/actions/git_clone_and_push.rb
jamesrb1: Done!

PR #66, opened 2024-07-22
Add links to versions and billing client in hybrid VERSIONS.md

tonidero:
To make it easier to navigate, we want to add links to the releases in the `VERSIONS.md` file of the hybrids + what version of the billing client it uses. This PR adds both of those, so it also auto-migrates existing VERSIONS.md files with the new column

After merging this, we need to fix the `purchases-unity` job since that one already has an extra column. Just updating the `VERSIONS.md` to have the new column should be enough.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
@@ -29,8 +29,8 @@ module Fastlane
           extend_header = lines[0].split('|').length <= 6
 
           if extend_header
-            lines[0] = "#{lines[0].strip} Android Billing Client version |\n"
-            lines[1] = "#{lines[1].strip}--------------------------------|\n"
+            lines[0] = "#{lines[0].strip} Play Billing Library version |\n"
+            lines[1] = "#{lines[1].strip}------------------------------|\n"
             lines.each_with_index do |line, index|
               if index > 1
                 lines[index] = "#{line.strip} |\n"
--- a/lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
@@ -24,7 +24,11 @@ module Fastlane
 
         File.open(versions_file_path, 'r+') do |file|
           lines = file.each_line.to_a
-          if lines[0].split('|').length <= 6
+          # Determine if the header needs to be extended based on the number of columns
+          # The number 6 comes from new_sdk_version, ios_version, android_version, hybrid_common_version + separators before and after
+          extend_header = lines[0].split('|').length <= 6
+
+          if extend_header
             lines[0] = "#{lines[0].strip} Android Billing Client version |\n"
             lines[1] = "#{lines[1].strip}--------------------------------|\n"
             lines.each_with_index do |line, index|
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
```diff
@@ -19,9 +19,25 @@ def self.run(params)
         ios_version = Helper::UpdateHybridsVersionsFileHelper.get_ios_version_for_hybrid_common_version(hybrid_common_version)
         UI.message("Obtained ios version #{ios_version} for PHC version #{hybrid_common_version}")
 
+        billing_client_version = Helper::UpdateHybridsVersionsFileHelper.get_android_billing_client_version(android_version)
+        UI.message("Obtained android billing client version #{billing_client_version} for PHC version #{hybrid_common_version}")
+
         File.open(versions_file_path, 'r+') do |file|
           lines = file.each_line.to_a
-          lines.insert(2, "| #{new_sdk_version} | #{ios_version} | #{android_version} | #{hybrid_common_version} |\n")
+          if lines[0].split('|').length <= 6
```
aboedo, line 27: super magical number here, maybe there's a way to clarify what it means? it was easier to understand before. maybe extracting a boolean + explaining why "6" makes sense, 

i.e: it means we get a column for each one of these: new_sdk_version, ios_version, android_version, hybrid_common_version, billing_client_version + the separator before and after

```ruby
# Determine if the header needs to be extended based on the number of columns
  # The number 6 comes from new_sdk_version, ios_version, android_version, hybrid_common_version, billing_client_version + separators before and after
  extend_header = lines[0].split('|').length <= 6

  if extend_header
```

lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
```diff
@@ -19,9 +19,25 @@ def self.run(params)
...
+              if index > 1
+                lines[index] = "#{line.strip} |\n"
+              end
+            end
+          end
+          lines.insert(2, "| #{new_sdk_version} " \
+                          "| [#{ios_version}](https://github.com/RevenueCat/purchases-ios/releases/tag/#{ios_version}) " \
+                          "| [#{android_version}](https://github.com/RevenueCat/purchases-android/releases/tag/#{android_version}) " \
+                          "| [#{hybrid_common_version}](https://github.com/RevenueCat/purchases-hybrid-common/releases/tag/#{hybrid_common_version}) " \
+                          "| [#{billing_client_version}](https://developer.android.com/google/play/billing/release-notes) |\n")
```
aboedo, line 40: I think you can avoid explicitly adding the ` | ` separator by using `Array::join`, like: 

```ruby
new_line = [
    new_sdk_version,
    "[#{ios_version}](https://github.com/RevenueCat/purchases-ios/releases/tag/#{ios_version})",
    "[#{android_version}](https://github.com/RevenueCat/purchases-android/releases/tag/#{android_version})",
    "[#{hybrid_common_version}](https://github.com/RevenueCat/purchases-hybrid-common/releases/tag/#{hybrid_common_version})",
    "[#{billing_client_version}](https://developer.android.com/google/play/billing/release-notes)"
  ].join(' | ')
```

aboedo: can probably use it for the previous section too

aboedo (commented):
looks great! left a couple of minor comments

lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
tonidero: That's... a totally good point 😅 I wanted to extract to a constant before and I forgot. But adding a comment also works 👍

lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
```diff
@@ -19,9 +19,25 @@ def self.run(params)
...
         UI.message("Obtained ios version #{ios_version} for PHC version #{hybrid_common_version}")
 
+        billing_client_version = Helper::UpdateHybridsVersionsFileHelper.get_android_billing_client_version(android_version)
+        UI.message("Obtained android billing client version #{billing_client_version} for PHC version #{hybrid_common_version}")
+
         File.open(versions_file_path, 'r+') do |file|
           lines = file.each_line.to_a
-          lines.insert(2, "| #{new_sdk_version} | #{ios_version} | #{android_version} | #{hybrid_common_version} |\n")
+          if lines[0].split('|').length <= 6
+            lines[0] = "#{lines[0].strip} Android Billing Client version |\n"
```
JayShortway, line 32: Nit: the official name Google uses these days is Play Billing Library. Maybe we should use the same name externally?

lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
```diff
@@ -19,9 +19,25 @@ def self.run(params)
...
+            lines[1] = "#{lines[1].strip}--------------------------------|\n"
```
JayShortway, line 33: Does it make sense to add the StoreKit version(s) as well? (I suppose a single SDK version can support one or multiple StoreKit versions, so that might make things more complicated.)

JayShortway (approved):
Thanks for taking care of this so quickly! Just a suggestion and a question.

lib/fastlane/plugin/revenuecat_internal/actions/update_hybrids_versions_file_action.rb
tonidero: Yeah... with version 5.0 of the iOS sdk we switched the default to StoreKit 2, but both should still be supported, so I don't think it provides as much value IMO. But we can add it later if needed!

JayShortway: Yup, sounds reasonable!

PR #67, opened 2024-08-01
Supports version numbers with build metadata

JayShortway:
## What
This PR adds support for version numbers with build metadata. [semver.org](https://semver.org/#spec-item-10)'s definition of build metadata is: 
>Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3----117B344092BD.

The following code paths have been updated to support build metadata: 
- `determine_next_version_using_labels()`
- `bump_version_update_changelog_create_pr()`

I think that covers everything, but let me know if you think of any other places that need updating!

It's quite a big diff, but the majority is tests. Did what I could to minimize this without making the tests hard to read. Let me know if you see other opportunities. 

## Why
We intend to use this to be able to specify the PHC version as build metadata for purchases-kmp, to make it easier for developers to figure out which PHC version they need to link against.


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -161,18 +161,91 @@ def self.detect_bump_type(version_name_a, version_name_b)
...
+        if append_hybrid_common_version
+          UI.user_error!("Appending the PHC version to prerelease versions violates SemVer.") if include_prereleases
+          UI.user_error!("Cannot append a nil PHC version.") if hybrid_common_version.nil?
+          UI.user_error!("Cannot append a blank PHC version.") if hybrid_common_version.strip.empty?
+        end
+      end
+
+      def self.validate_new_version_if_appending_phc_version?(append_hybrid_common_version, new_version_number, hybrid_common_version)
+        if append_hybrid_common_version
+          if new_version_number.include?("-")
```
tonidero, line 174: I've been thinking, would it be cleaner to separate these into a constant? So something like: `PRERELEASE_DELIMITER = "-"`, same for metadata delimiter.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -161,18 +161,91 @@ def self.detect_bump_type(version_name_a, version_name_b)
...
+        elsif should_append_phc_version?(append_hybrid_common_version, include_prereleases, hybrid_common_version, new_version_number)
+          UI.important(
+            "Appending PHC version (+#{hybrid_common_version}) to new version (#{new_version_number}), as instructed."
+          )
+          return "#{new_version_number}+#{hybrid_common_version}"
+        end
+        return new_version_number
+      end
+
+      private_class_method def self.should_ask_to_append_phc_version?(append_hybrid_common_version, include_prereleases, hybrid_common_version, new_version_number)
```
tonidero, line 201: Hmm I'm thinking what the use case for this one is? This will only be asked if done locally right? In that case, I think we should just try to imitate what we do in CI? Maybe I'm missing a case we would want something different locally though...

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -161,18 +161,91 @@ def self.detect_bump_type(version_name_a, version_name_b)
...
+        UI.interactive? &&
+        append_hybrid_common_version.nil? &&
```
vegaro, line 203: should this check for `nil` or should it check the value of `append_hybrid_common_version`?

It feels like these two private methods `should_ask_to_append_phc_version` and `should_append_phc_version` could be combined into one and then check if the UI is interactive to show a `UI.confirm` or a `UI.important`. But maybe I am missing something

spec/helper/versioning_helper_spec.rb
```diff
@@ -794,6 +860,282 @@ def mock_native_releases
...
+        append_on_confirmation: false,
+        append_hybrid_common_version: nil,
+        include_prereleases: false,
+        hybrid_common_version: "12.0.0",
+        new_version_number: "2.0.0+",
+        expected_version: "2.0.0+"
+      )
+    ]
+
+    params.each_with_index do |param, i|
```
tonidero, line 1119: Nitpick but wdyt about calling the `params` array something like `params_sets` and then calling `param` `params` instead?

spec/helper/versioning_helper_spec.rb
```diff
@@ -794,6 +860,282 @@ def mock_native_releases
     end
   end
 
+  describe 'append_phc_version_if_necessary' do
+    params = [
```
tonidero, line 864: ❤️

spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
```diff
@@ -202,6 +202,421 @@
       )
     end
 
+    it 'fails trying to append PHC version if is_prerelease is true' do
+      # Arrange
+      new_version_provided = new_version
+      is_prerelease = true
+
+      # Act, Assert
```
tonidero, line 210: These tests are nice but I'm wondering if it's worth delimiting them with these comments... Not a strong opinion though, I'm ok leaving as is.

tonidero (approved):
Nice! This LGTM!

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
JayShortway: The idea here was to minimize the chance of errors when releasing locally (e.g. majors). Without this you'd need to remember to append the PHC version when releasing purchases-kmp. Now, if you input a version number without the build metadata, it asks if you want to append the PHC version as build metadata. 

I guess an alternative would be to hard code a list of repos for which we always want to append the PHC version?

spec/actions/bump_version_update_changelog_create_pr_action_spec.rb
JayShortway: No strong opinion either, removed! https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/67/commits/07db6a548a71e9f1e86bacdf24933dfa28ab5a95

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
tonidero: Hmm but the repos using PHC as metadata would already define the `append_hybrid_common_version` parameter to `true` right? So I think it's ok to assume if you're bumping the version in a repo where that's defined without passing the metadata, just append the PHC version. Wdyt?

JayShortway: I see what you mean. I think we can make that work with a slight adjustment. `append_hybrid_common_version` should be something like `append_hybrid_common_version_if_next_version_not_prerelease` (and behave in that way too). 
 
The reason why I'm thinking that is that we could call `bump` (manually) without specifying a version (yet). If we then input a prerelease version (when the CLI asks for it), and `append_hybrid_common_version` is hardcoded to true in the repo, that would be an invalid state. (We currently throw an error.) However we should be able to release prerelease versions via the CLI of course.

Hope it's somewhat clear. What do you think?

tonidero: Ahhh yes I understand the issue now. And yeah I think your proposed solution should work 👍

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -161,18 +161,91 @@ def self.detect_bump_type(version_name_a, version_name_b)
         end
       end
 
+      def self.validate_input_if_appending_phc_version?(append_hybrid_common_version, include_prereleases, hybrid_common_version)
+        if append_hybrid_common_version
+          UI.user_error!("Appending the PHC version to prerelease versions violates SemVer.") if include_prereleases
+          UI.user_error!("Cannot append a nil PHC version.") if hybrid_common_version.nil?
+          UI.user_error!("Cannot append a blank PHC version.") if hybrid_common_version.strip.empty?
```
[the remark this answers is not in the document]
tonidero: Ohh I think that was a rails method: https://apidock.com/rails/Object/blank%3F. Please ignore me then 🙏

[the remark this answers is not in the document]
JayShortway: `blank?` seems to be a Rails thing as far as I can tell?

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
JayShortway: It was checking for `nil`, because that meant no decision had been made about whether we should append the PHC version or not, and so we should ask the user.

However this logic is now removed. We never ask the user anymore. Repos that need this will have the relevant parameter hardcoded to true. See also this comment thread: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/67#discussion_r1700360264.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
JayShortway: Good one! https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/67/commits/c8e582375131f5e18667369dd5997278e6064f2e

spec/helper/versioning_helper_spec.rb
JayShortway: Yea I like it! https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/67/commits/1a729a73806028dfa996a1ec54769966819a6838

spec/helper/versioning_helper_spec.rb
JayShortway: 🫶

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
JayShortway: This is now done! We never ask the user anymore. Repos that need to append the PHC version will have `append_phc_version_if_next_version_is_not_prerelease` hardcoded to true. 

Had to adjust some tests for this, and could remove some code. If `append_phc_version_if_next_version_is_not_prerelease` is true, but the logic decides _not_ to append, the reason for that decision is printed to std out. (See `VersioningHelper.should_append_phc_version?()`.)

PR #68, opened 2024-08-18
New `scan_with_flaky_test_retries` action

joshdholtz:
## Motivation

Deal with flaky tests one and for all 💪 

Easy way to rerun failed tests that are failing to do unresetable state (like SKTest or integration tests) in a whole new `scan` (`xcodebuild`) run

## Description

New `scan_with_flaky_test_retries` action

- Replaces custom `retry_failed_tests` lane from iOS repo (based off of that logic)
  - See https://github.com/RevenueCat/purchases-ios/pull/4188 for how this is being used
- Takes same options as `scan` (using `Scan::Options.available_options`)
  - But with an additional `number_of_flaky_retries` option

### Logic
1. Creates a new temporary directory for each run of the tests
2. Runs scan almost the exact same parameters
    - Replaces `fail_build`, `only_testing`, and `output_directory`
3. After test are run, inspects the junit file for any failed tests. If there are any failed tests...
    - Writes all the failed tests to a file to retry again
    - Merges the junit report with the original (if it was a retry)
    - Repeats this step until `number_of_flaky_retries` has been reached or no more failed tests
4. Moves all the artifacts (xcresult, merged junit file, and all individual junit files) into the original `output_directory` 

If `number_of_flaky_retries` is 0, it will just pipe everything to `scan` (so that it will run without any other side effects)

#### Before

```rb
scan(
  step_name: "scan - iPhone",
  device: ENV['SCAN_DEVICE'] || "iPhone 12 (15.2)",
  ensure_devices_found: true,
  scheme: "RevenueCat",
  prelaunch_simulator: true,
  output_types: 'junit',
  result_bundle: true,
  testplan: test_plan,
  configuration: 'Debug',
  output_directory: "fastlane/test_output/xctest/ios",
  number_of_retries: generate_snapshots ? 0 : 5, # Number of retries in an xcodebuild run
  xcargs: "OTHER_SWIFT_FLAGS='-DCUSTOMER_CENTER_ENABLED'", # todo: remove once Customer Center is released,
)
```

#### After

```rb
scan_with_flaky_test_retries(     # <---------- THIS IS JUST A SIMPLE RENAME
  step_name: "scan - iPhone",
  device: ENV['SCAN_DEVICE'] || "iPhone 12 (15.2)",
  ensure_devices_found: true,
  scheme: "RevenueCat",
  prelaunch_simulator: true,
  output_types: 'junit',
  result_bundle: true,
  testplan: test_plan,
  configuration: 'Debug',
  output_directory: "fastlane/test_output/xctest/ios",
  number_of_retries: generate_snapshots ? 0 : 5, # Number of retries in an xcodebuild run
  xcargs: "OTHER_SWIFT_FLAGS='-DCUSTOMER_CENTER_ENABLED'", # todo: remove once Customer Center is released,

  number_of_flaky_retries: generate_snapshots ? 0 : 5 # Number of retries in an xcodebuild run
  # ^^^^^^^ Just one new option to add
)
```


jamesrb1 (approved):
This will be so nice! 💖
Thank you!

lib/fastlane/plugin/revenuecat_internal/actions/scan_with_flaky_test_retries.rb
```diff
@@ -0,0 +1,261 @@
...
+        if number_of_flaky_retries == 0
+          # Run scan directly if no retries are needed
+          other_action.scan(**param_values)
+        else
+          # Delete option that is part of scan but we will replace
+          output_directory = param_values.delete(:output_directory)
+
+          # Storing all scan output in a separate temp directory
+          begin
+            temp_dir = Dir.mktmpdir
```
fire-at-will, line 27: Nit: It doesn't look like this directory is ever cleaned up. I _assume_ that CircleCI will probably delete it when the job is finished and it destroys the image. 

Would it be a good idea to delete it when we're done in case this ever gets run on someone's local machine or in a CI environment that persists files across CI runs?

test_projects/.DS_Store
fire-at-will, line 1: Can you remove the .DS_Stores and add them to the `.gitignore`?

test_projects/FlakyTests/FlakyTestsTests/FlakyTestsTests.swift
```diff
@@ -0,0 +1,35 @@
...
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
+
+    func testRandomFailure() {
+        // Generate a random number between 0 and 1
+        let randomValue = Double.random(in: 0..<1)
+
+        // Check if the random value falls within the 75% failure range
+        if randomValue < 0.75 {
```
fire-at-will, line 26: This is my favorite unit test ever written 🤩😂

fire-at-will (approved):
Very nice, I can't wait to start using it!! I had a few nitpicks/questions but looks good otherwise! Thanks for taking this on 👏

PR #69, opened 2024-09-04
Add support for labels with `pr:` prefix

vegaro:
We are going to add `pr:` as a prefix for Pull Request labels. This PR adds support for both versions, while we transition all repos to use `pr:` labels

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
@@ -8,7 +8,7 @@ module Fastlane
 
   module Helper
     class GitHubHelper
-      SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf revenuecatui refactor style test next_release dependencies phc_dependencies minor].to_set
+      SUPPORTED_PR_LABELS = (%w[breaking build ci docs feat fix perf revenuecatui refactor style test next_release dependencies phc_dependencies minor].map { |label| "pr:#{label}" }).to_set
 
       def self.get_pr_resp_items_for_sha(sha, github_token, rate_limit_sleep, repo_name, base_branch)
         if rate_limit_sleep > 0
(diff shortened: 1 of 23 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
```diff
@@ -8,7 +8,9 @@ module Fastlane
 
   module Helper
     class GitHubHelper
-      SUPPORTED_PR_LABELS = %w[breaking build ci docs feat fix perf revenuecatui refactor style test next_release dependencies phc_dependencies minor].to_set
+      PR_LABEL_PREFIX = 'pr:'
+      BASE_LABELS = %w[breaking build ci docs feat fix perf revenuecatui refactor style test next_release dependencies phc_dependencies minor]
+      SUPPORTED_PR_LABELS = (BASE_LABELS + BASE_LABELS.map { |label| "#{PR_LABEL_PREFIX}#{label}" }).to_set
```
tonidero, line 13: Hmm I've been thinking whether it makes sense to just remove support for the old labels now, and just update each SDK to the new version when the labels have been updated... Wdyt?

vegaro: Right. Good point. I'm over complicating things. With the dangerfile we can't do that, but we can for the fastlane plugin

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -264,17 +264,17 @@ def self.append_phc_version_if_necessary(append_phc_version_if_next_version_is_n
 
       # rubocop:disable Metrics/PerceivedComplexity
       private_class_method def self.get_section_depending_on_types_of_change(change_types)
-        if change_types.include?("breaking")
+        if change_types.any?("pr:breaking")
```
tonidero, line 267: Hmm why the change from `include?` to `any?`

tonidero (approved):
Just a question but LGTM otherwise

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: oopsies, this is from before, where it could be with or without prefix. Good catch

tonidero (approved):

PR #70, opened 2024-09-06
Add features sections to changelog and emojis

vegaro:
- The changelog is now split between `RevenueCat SDK` and `RevenueCatUI SDK`. Something tagged with `pr:revenuecatui` will go into the `RevenueCatUI` section
- There can be arbitrary feature labels now. For example prs with labels `feat:Customer Center` or `feat:Paywall Components` will be rendered under new sections.
- These new feature sections have subsections for fixes and new features. So a PR should still be tagged with the `pr:` label (for example `feat:Customer Center` and `pr:fix`. If the PR label is not `pr:feat` or `pr:fix` go directly to `Other changes` section (outside of the specific functionality). This is to keep “internal” features in `Other changes`, making the changelog more readable for developers
- I unified the labels: `pr:docs`, `pr:perf` to `pr:fix`. And `pr:build`, `pr:ci`, `pr:refactor`, `pr:style`, `pr:test` to `pr:other`
- If a release only contains `pr:other` prs, it will be skipped.
- Renamed `pr:minor` to `pr:force_minor`. Created `pr:force_patch` to be able to force releases with only `pr:other`

```
## RevenueCat SDK
### ✨ New Features
* added a log when `autoSyncPurchases` is disabled (#1749) via aboedo (@aboedo)

## RevenueCatUI SDK
### Customer Center
#### ✨ New Features
* `Customer Center`: contact support (#2949) via aboedo (@aboedo)
#### 🐞 Bugfixes
* `Customer Center`: a fix (#2949) via aboedo (@aboedo)
### Paywall Components
#### ✨ New Features
* `Paywalls Components`: this is amazing (#2949) via aboedo (@aboedo)

### 🔄 Other Changes
* `Paywalls Components`: another amazing thing (#2949) via aboedo (@aboedo)
```

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
@@ -16,9 +16,9 @@ module Fastlane
 
   BUMP_PER_LABEL = {
     major: %w[pr:breaking].to_set,
-    minor: %w[pr:feat pr:minor].to_set,
-    patch: %w[pr:docs pr:fix pr:perf pr:dependencies pr:phc_dependencies pr:revenuecatui].to_set,
-    skip: %w[pr:build pr:ci pr:refactor pr:style pr:test pr:next_release].to_set
+    minor: %w[pr:feat pr:force_minor].to_set,
+    patch: %w[pr:fix pr:dependencies pr:phc_dependencies pr:force_patch].to_set,
+    skip: %w[pr:other pr:next_release].to_set
   }
 
   module Helper
@@ -51,7 +51,14 @@ module Fastlane
 
         commits = Helper::GitHubHelper.get_commits_since_old_version(github_token, old_version, repo_name)
 
-        changelog_sections = { breaking_changes: [], new_features: [], paywalls: [], fixes: [], performance: [], dependency_updates: [], other: [], work_in_progress: {} }
+        changelog_sections = {
+          breaking_changes: [],
+          new_features: [],
+          fixes: [],
+          dependency_updates: [],
+          features: {},
+          other: []
+        }
 
         commits.map do |commit|
           name = commit["commit"]["author"]["name"]
@@ -74,7 +81,7 @@ module Fastlane
               # Append links to native releases
               line += native_releases_links(github_token, hybrid_common_version, versions_file_path)
             end
-            if section == :work_in_progress
+            if section == :features
               changelog_sections[section][subsection] ||= []
               changelog_sections[section][subsection].push(line)
             else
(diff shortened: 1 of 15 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -238,33 +242,47 @@ def self.append_phc_version_if_necessary(append_phc_version_if_next_version_is_n
...
+            title = "### ⚡ Performance Improvements"
           when :dependency_updates
-            title = "### Dependency Updates"
+            title = "### 📦 Dependency Updates"
+          when :work_in_progress
+            title = "### 🚧 Work in Progress"
+            wip_sections = prs.map do |subsection, items|
+              capitalized_subsection = subsection.split.map(&:capitalize).join(' ')
+              "#### #{capitalized_subsection}\n#{items.join("\n")}"
+            end
```
joshdholtz, line 268: :chefs_kiss:

joshdholtz (approved):
This looks good to me! Thanks for adding this 😁 :shipit:

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -238,33 +242,47 @@ def self.append_phc_version_if_necessary(append_phc_version_if_next_version_is_n
...
+            lines = prs
+          end
+          "#{title}\n#{lines.join("\n")}"
         end.join("\n")
       end
 
-      # rubocop:disable Metrics/PerceivedComplexity
       private_class_method def self.get_section_depending_on_types_of_change(change_types)
-        if change_types.include?("pr:breaking")
+        wip_label = change_types.find { |type| type.end_with?(":wip") }
```
vegaro, line 282: main reason for it to being a suffix :wip  is that I think we can add another prefix feat: different than pr: for bigger arbitrary features if we want, but that can come in a different future PR since that requires a bigger refactor

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -51,7 +51,7 @@ def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep, incl
 
         commits = Helper::GitHubHelper.get_commits_since_old_version(github_token, old_version, repo_name)
 
-        changelog_sections = { breaking_changes: [], new_features: [], paywalls: [], fixes: [], performance: [], dependency_updates: [], other: [] }
+        changelog_sections = { breaking_changes: [], new_features: [], paywalls: [], fixes: [], performance: [], dependency_updates: [], other: [], work_in_progress: {} }
```
tonidero, line 54: This might be just my opinion, but I don't think we should have a "WIP" section, since it doesn't provide much value to customer looking at the changelog and for us, we could just have them in an "Other" or "internal" section.

I think having custom sections (like Paywalls/Customer center) makes more sense for public facing sections of the changelog, which "WIP" would not be.

This might just be my opinion though. Happy to chat more about this over Slack!

jamesrb1: I agree these WIP doesn't hold value for most customers most of the time. Maybe we could put these inside [a toggle](https://gist.github.com/scmx/eca72d44afee0113ceb0349dd54a84a2) for us or others who may find them useful?

vegaro: those are valid points. I was also thinking of adding `feat:` as a prefix for arbitary features. For example `feat: Customer Center`, or `feat: Paywall Components`. Then things would be nested using the same sections we normally use (new features, fixes, other...). What do you think?

tonidero: Yeah, I think this makes total sense for `feat:` or `fix:`, but not for `wip:` IMO... I don't think we need a "pretty" changelog for ourselves, and customers won't care about anything in `wip:`. 

I would say that the changelog should mostly consist of 
- Breaking
- Feats (with any subsections per product)
- Bugfixes (with any subsections per product)
- Other (basically everything that the developer doesn't care about and/or is not user facing)

Again, happy to chat more about this though!

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -297,7 +337,7 @@ def self.append_phc_version_if_necessary(append_phc_version_if_next_version_is_n
       private_class_method def self.get_type_of_change_from_pr_info(pr_info)
         pr_info["labels"]
           .map { |label_info| label_info["name"].downcase }
-          .select { |label| Helper::GitHubHelper::SUPPORTED_PR_LABELS.include?(label) }
+          .select { |label| Helper::GitHubHelper::SUPPORTED_PR_LABELS.include?(label) || label.start_with?("feat:") }
```
tonidero, line 340: If we do rename `features` to something else, we might also want to rename these labels

tonidero (commented):
LGTM, just some questions over Slack of the structure + the naming for features vs new features being confusing.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -16,12 +16,13 @@ module Fastlane
 
   BUMP_PER_LABEL = {
     major: %w[pr:breaking].to_set,
-    minor: %w[pr:feat pr:minor].to_set,
-    patch: %w[pr:docs pr:fix pr:perf pr:dependencies pr:phc_dependencies pr:revenuecatui].to_set,
-    skip: %w[pr:build pr:ci pr:refactor pr:style pr:test pr:next_release].to_set
+    minor: %w[pr:feat pr:force_minor].to_set,
+    patch: %w[pr:fix pr:dependencies pr:phc_dependencies pr:force_patch].to_set,
+    skip: %w[pr:other pr:next_release].to_set
```
tonidero, line 21: We will need to update the `Dangerfile` with these new/old labels

tonidero (approved):
LGTM! We can see how this works and iterate from that if needed 👍

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
vegaro: https://github.com/RevenueCat/Dangerfile/pull/10

PR #71, opened 2024-09-26
Fixes bump detection for pr:force_patch and pr:force_minor

JayShortway:
As the title says. I noticed that `pr:force_minor` wasn't working in https://github.com/RevenueCat/purchases-kmp/pull/219#issuecomment-2376221971. 
  
Added tests to verify `pr:force_minor` and `pr:force_patch`. The fix was to add those labels to `SUPPORTED_PR_LABELS`.


tonidero (approved):
Ahh right, good catch! Thanks for fixing it! 👍

PR #72, opened 2024-10-02
Fix changelog generation for hybrids to include native release links

tonidero:
This broke in #66. We should parse markdown links for the releases correctly after this change.


JayShortway (approved):
LGTM!

PR #73, opened 2024-12-12
Fix changelog generation when there are commits without PRs

tonidero:
From some tests, the `author` field in the Github response is null on commits that were merged without a PR. This was causing the automation to fail with a 
```
versioning_helper.rb:113:in `block in auto_generate_changelog': \e[31m[!] undefined method `[]' for nil:NilClass\e[0m (NoMethodError)
``` 

error.

I didn't see another way to get the username, so I just removed it from the changelog for now

### DIFF
--- a/spec/helper/versioning_helper_spec.rb
+++ b/spec/helper/versioning_helper_spec.rb
@@ -447,7 +447,7 @@ describe Fastlane::Helper::VersioningHelper do
         nil
       )
       expect(changelog).to eq("### 🔄 Other Changes\n" \
-                              "* Updating great support link via Miguel José Carranza Guisado (@MiguelCarranza)")
+                              "* Updating great support link via Miguel José Carranza Guisado")
     end
 
     def mock_native_releases
--- a/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
@@ -110,8 +110,7 @@ module Fastlane
             UI.important("Cannot find pull request associated to #{sha}. Using commit information and adding it to the Other section")
             message = commit["commit"]["message"]
             name = commit["commit"]["author"]["name"]
-            username = commit["author"]["login"]
-            line = "* #{message} via #{name} (@#{username})"
+            line = "* #{message} via #{name}"
             changelog_sections[:other].push(line)
           else
             UI.user_error!("Cannot generate changelog. Multiple commits found for #{sha}")


spec/helper/versioning_helper_spec.rb
```diff
@@ -447,7 +447,7 @@
         nil
       )
       expect(changelog).to eq("### 🔄 Other Changes\n" \
-                              "* Updating great support link via Miguel José Carranza Guisado (@MiguelCarranza)")
```
tonidero, line 450: Seems like this was working before 🤔... Not sure if the github API changed 😕

nicfix (approved):
Approving to unlock our pipeline, however I think we might want to investigate more on trying to get the username back.

JayShortway (approved):
Approved! Do we have examples of which commits it failed on?

tonidero:
This is one of the jobs where it failed: https://app.circleci.com/pipelines/github/RevenueCat/purchases-js/1067/workflows/bc31318f-a3c5-4714-9d90-5239d976d10c/jobs/2993. It failed due to this commit: https://github.com/RevenueCat/purchases-js/commit/6089d084620d42c0b9f995b75210505dc874acff

Investigating, we do the request to: https://api.github.com/repos/RevenueCat/purchases-js/compare/2bd4d22c7ca48dba9225ac1632e22e69bab937b2...9a2f2b7b7cb58f4d857a2510d0d3d79872f54239 and for that commit, the `author` field is null: 
```
{
    "sha": "6089d084620d42c0b9f995b75210505dc874acff",
    "node_id": "C_kwDOKlHSu9oAKDYwODlkMDg0NjIwZDQyYzBiOWY5OTViNzUyMTA1MDVkYzg3NGFjZmY",
    "commit": {
        ....
    },
    "url": "https://api.github.com/repos/RevenueCat/purchases-js/commits/6089d084620d42c0b9f995b75210505dc874acff",
    "html_url": "https://github.com/RevenueCat/purchases-js/commit/6089d084620d42c0b9f995b75210505dc874acff",
    "comments_url": "https://api.github.com/repos/RevenueCat/purchases-js/commits/6089d084620d42c0b9f995b75210505dc874acff/comments",
    "author": null,
    "committer": null,
    "parents": [
        ....
    ]
},
```

tonidero:
> however I think we might want to investigate more on trying to get the username back.

Note that this change affects only commits without an associated PR, so we should hopefully never find ourselves on this situation... We can dedicate more time to find options for sure, but not sure it's worth it right now.

JayShortway:
Interesting that the Guido's name also can't be clicked in that commit you linked. (Normally you can click the author's name to go to their GitHub profile.) Maybe merged from the command line with a different `user.email` set?

Definitely not needed to find other options imo. Just curious 😇

PR #74, opened 2025-03-03
[DX-287] Temporarily forks the fastlane's `SetGithubReleaseAction` to allow support for `make_latest`

polpielladev:
We have recently realised that [fastlane's `set_github_release`](https://docs.fastlane.tools/actions/set_github_release/) does not support passing the `make_latest` parameter to the body of the request to GitHub's API.

To address this in the quickest way possible, let's fork the implemenation and add support for the new field. This new addition means that we can support marking only the latest stable release as the latest in GitHub.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
@@ -5,9 +5,9 @@ require 'fastlane/actions/push_to_git_remote'
 require 'fastlane/actions/create_pull_request'
 require 'fastlane/actions/ensure_git_branch'
 require 'fastlane/actions/ensure_git_status_clean'
-require 'fastlane/actions/set_github_release'
 require 'fastlane/actions/reset_git_repo'
 require_relative 'versioning_helper'
+require_relative 'github_helper'
 
 module Fastlane
   UI = FastlaneCore::UI unless Fastlane.const_defined?(:UI)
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -145,8 +145,9 @@ def self.calculate_next_snapshot_version(current_version)
       def self.create_github_release(release_version, release_description, upload_assets, repo_name, github_api_token)
         commit_hash = Actions.last_git_commit_dict[:commit_hash]
         is_prerelease = release_version.include?(DELIMITER_PRERELEASE)
+        is_latest_stable_release = !is_prerelease && newer_than_latest_published_version?(release_version)
 
-        Actions::SetGithubReleaseAction.run(
+        Helper::GitHubHelper.create_github_release(
```
tonidero, line 150: Might be worth adding a comment here as well, so we remember to revert this once it's supported in Fastlane.

tonidero (approved):
I think this makes sense as a temporal workaround! Thank you! Also cc @joshdholtz, since it might be something worth moving to fastlane at some point :)

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
polpielladev: Done! ✅

polpielladev:
@joshdholtz @tonidero I had a bit of time today before signing off, so I decided to upstream these changes and bring them to fastlane.

✨  Here is the PR: https://github.com/fastlane/fastlane/pull/29475 ✨

PR #76, opened 2025-03-06
feat: Add exponential backoff retry for pod_push

facumenzella:
This PR adds an exponential backoff retrial strategy for pod_push 🐇 

Example output
```
[16:55:53]: ▸ Updating spec repo `cocoapods`
[16:55:53]: ▸ Validating podspec
[16:58:17]: ▸  -> RevenueCat (5.19.0)
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Using codesigning identity override: -
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Building targets in dependency order
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Target dependency graph (4 targets)
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Metadata extraction skipped. No AppIntents.framework dependency found. (in target 'RevenueCat' from project 'Pods')
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Signing static framework with --generate-pre-encrypt-hashes (in target 'Pods-App' from project 'Pods')
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Metadata extraction skipped. No AppIntents.framework dependency found. (in target 'App' from project 'App')
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Disabling previews because SWIFT_VERSION is set and SWIFT_OPTIMIZATION_LEVEL=-O, expected -Onone (in target 'App' from project 'App')
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Disabling previews because SWIFT_VERSION is set and SWIFT_OPTIMIZATION_LEVEL=-O, expected -Onone (in target 'Pods-App' from project 'Pods')
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Disabling previews because SWIFT_VERSION is set and SWIFT_OPTIMIZATION_LEVEL=-O, expected -Onone (in target 'RevenueCat' from project 'Pods')
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Disabling previews because SWIFT_VERSION is set and SWIFT_OPTIMIZATION_LEVEL=-O, expected -Onone (in target 'RevenueCat-RevenueCat' from project 'Pods')
[16:58:17]: ▸ - NOTE  | xcodebuild:  note: Using codesigning identity override:
[16:58:17]: ▸ [!] Calling the GitHub commit API timed out.
[...]

PR #77, opened 2025-03-21
feat: Retry on internal server error for pod_repo_push

facumenzella:
Follow up of https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/76

We're also retrying on internal server error 🚀

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,5 +2,7 @@ source('https://rubygems.org')
 
 gemspec
 
+gem 'abbrev'
+
 plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
 eval_gemfile(plugins_path) if File.exist?(plugins_path)
--- a/lib/fastlane/plugin/revenuecat_internal/actions/pod_push_with_error_handling.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/pod_push_with_error_handling.rb
@@ -2,37 +2,61 @@ require 'fastlane/action'
 require 'fastlane_core/ui/ui'
 require 'fastlane_core/configuration/config_item'
 require 'fastlane/actions/pod_push'
-require_relative '../helper/revenuecat_internal_helper'
 
 module Fastlane
   module Actions
     class PodPushUnknownError < StandardError; end
 
     class PodPushWithErrorHandlingAction < Action
+      MAX_RETRIES = 3
+      INITIAL_DELAY = 5
+
       def self.run(params)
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -2,5 +2,7 @@ source('https://rubygems.org')
 
 gemspec
 
+gem 'abbrev'
```
tonidero, line 5: Hmm is this added for any reason? Couldn't find anything that uses it.

lib/fastlane/plugin/revenuecat_internal/actions/pod_push_with_error_handling.rb
```diff
@@ -30,19 +30,23 @@ def self.run(params)
             output_str = e.message
 
             if output_str.include?("[!] Unable to accept duplicate entry for:")
-              FastlaneCore::UI.error("⚠️ Duplicate entry detected. Skipping push.")
-              return true
+              UI.error("⚠️ Duplicate entry detected. Skipping push.")
+              return false
```
tonidero, line 34: why this change? Seems to be making some tests fail as well

tonidero (commented):
I believe some tests are failing but changes look good! Thanks for updating this so quickly! 💪

ajpallares (approved):
Looks good! ~Just one question~ it was the same as Toni's about `gem 'abbrev'`

Gemfile
tonidero: Hmm what's the error? Seems weird it's requiring this dependency without using it... If it's a dependency of a dependency, it shouldn't be part of the `Gemfile`, but just of the `Gemfile.lock` I think...

lib/fastlane/plugin/revenuecat_internal/actions/pod_push_with_error_handling.rb
facumenzella: I messed it up when fixing the merge conflict via GitHub. Fixed now :)

Gemfile
facumenzella:
```
warning: abbrev was loaded from the standard library, but is not part of the default gems starting from Ruby 3.4.0.
You can add abbrev to your Gemfile or gemspec to silence this warning.
```

tonidero: ohh interesting... I wonder if we had this warning before... I think this is probably fine, but I wonder if we need to add it to the gemspec as well... that would propagate the dependency to users of the plugin... But well, since it's a warning, i think we can do this for now and revisit later if needed 👍

PR #79, opened 2025-04-18
Only retry tests that failed but didn't succeed after in-run retry

joshdholtz:
## Motivation

The full retry of `scan` with only failed tests was retrying sometimes when it shouldn't. It was detecting failurs when tests were retried within Xcode that eventually succeeded.

## Description

Only writing failed tests if they don't have a successful run in the suite (this means they did eventually pass)

### Passing on first try

Screenshot 2025-04-18 at 1 43 26 PM

### Retrying after failures

| Failure | Successful retry |
|--------|--------|
| Screenshot 2025-04-18 at 1 53 53 PM | Screenshot 2025-04-18 at 1 54 13 PM |

### DIFF
--- a/fastlane-plugin-revenuecat_internal.gemspec
+++ b/fastlane-plugin-revenuecat_internal.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('rest-client')
 
   spec.add_development_dependency('bundler')
-  spec.add_development_dependency('fastlane')
+  spec.add_development_dependency('fastlane', '>= 2.227.1')
   spec.add_development_dependency('pry')
   spec.add_development_dependency('rake')
   spec.add_development_dependency('rspec')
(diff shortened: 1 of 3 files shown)


fastlane-plugin-revenuecat_internal.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('rest-client')
 
   spec.add_development_dependency('bundler')
-  spec.add_development_dependency('fastlane', '2.207.0')
+  spec.add_development_dependency('fastlane')
```
joshdholtz, line 27: Needed latest _fastlane_ due to fixes for Xcode 16.3

ajpallares: Even if we need to update `fastlane`, shouldn't we keep it pinned to a specific version?

ajpallares (approved):
I think this makes total sense! One small comment though, FFTI.

fastlane-plugin-revenuecat_internal.gemspec
joshdholtz: > Even if we need to update `fastlane`, shouldn't we keep it pinned to a specific version?

IMO, yes and no 🤷‍♂️ I caused issues for me being pinned to a specific version here. I can change this to be `>=` since there are some fixes needed in new version for `xcresulttool` so I will probably do that

PR #84, opened 2025-09-23
Consider `next_version` to create changelog

ajpallares:
When creating the Changelog, the current implementation only considers the changes from the latest tag to `HEAD`. However, when making a new release on a previous major means that the script will produce an invalid changelog, since the it will be generated from the latest tag (i.e. the latest major). 

This PR makes it so that the changelog is generated from the latest release **prior** to the `next_version`. 

For example, if we have a repo where the latest release is `17.7.0`, and we want to make a new minor release of the existing `14.2.0`, then the changelog for the new `14.3.0` release would consider the changes from `14.2.0` and not from `17.7.0`.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
@@ -48,7 +48,7 @@ module Fastlane
       def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep, include_prereleases, hybrid_common_version, versions_file_path, target_tag = nil)
         base_branch = Actions.git_branch
         Actions.sh("git fetch --tags -f")
-        old_version = latest_version_number_less_than(target_tag, include_prereleases: include_prereleases)
+        old_version = latest_version_number_smaller_than(target_tag, include_prereleases: include_prereleases)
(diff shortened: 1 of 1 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -45,10 +45,10 @@ def self.determine_next_version_using_labels(repo_name, github_token, rate_limit
       end
 
       # rubocop:disable Metrics/PerceivedComplexity
-      def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep, include_prereleases, hybrid_common_version, versions_file_path)
+      def self.auto_generate_changelog(repo_name, github_token, rate_limit_sleep, include_prereleases, hybrid_common_version, versions_file_path, target_tag = nil)
         base_branch = Actions.git_branch
         Actions.sh("git fetch --tags -f")
-        old_version = latest_version_number(include_prereleases: include_prereleases)
+        old_version = latest_version_number_less_than(target_tag, include_prereleases: include_prereleases)
```
vegaro, line 51: I am a bit confused about the `less_than` part in the name of the function

vegaro (approved):
ready to go after the rename

PR #85, opened 2025-09-24
Allow automatic bump in older majors

ajpallares:
Currently, when SDKs call `determine_next_version_using_labels`, this plugin always considers the latest version available. This works most of the time, i.e. for releases of the latest major.

However, sometimes we need to make new releases in older majors. In those scenarios, `determine_next_version_using_labels` was not useful because it would never consider older majors.

This PR adds an optional `current_version` (`nil` by default) parameter to `determine_next_version_using_labels` such that:
* If `nil`, `determine_next_version_using_labels` keeps the existing behavior: considers the latest version available.
* If not `nil`, `determine_next_version_using_labels` makes the calculation **only considering the releases for the same major as `current_version`**. This means that the returned next version will still be from the same major as `current_version`.


spec/helper/versioning_helper_spec.rb
```diff
@@ -148,7 +150,7 @@
     end
 
     it 'generates changelog automatically from github commits including feat section' do
-      setup_commit_search_stubs(hashes_to_responses_wip, get_commits_response_features)
+      setup_commit_search_stubs(hashes_to_responses_wip, get_commits_response_features, 'cfdd80f73d8c91121313d72227b4cbe283b57c1e')
```
ajpallares, line 153: Because I added the `tags` named parameter to `setup_commit_search_stubs`, if it kept the call like this
```rb
setup_commit_search_stubs(hashes_to_responses_wip, get_commits_response_features)
```

(according to ChatGPT) it seems Ruby 2.6 tries to interpret the `body` in `get_commits_response_features` as a parameter of `setup_commit_search_stubs`, raising an error like this:

```
1) FastlaneCore::Helper::VersioningHelper.auto_generate_changelog generates changelog automatically from github commits including feat section
     Failure/Error:
       def setup_commit_search_stubs(hashes_to_responses,
                                     commits_response = get_commits_response,
                                     last_release_sha = 'cfdd80f73d8c91121313d72227b4cbe283b57c1e',
                                     tags: ['0.1.0', '0.1.1', '1.11.0', '1.1.1.1', '1.1.1-alpha.1', '1.10.1'])
         setup_tag_stubs(tags: tags)
         mock_commits_since_last_release(last_release_sha, commits_response)
         hashes_to_responses.each do |hash, response|
           allow(Fastlane::Actions::GithubApiAction).to receive(:run)
             .with(server_url: server_url,
                   path: "/search/issues?q=repo:RevenueCat/mock-repo-name+is:pr+base:main+SHA:#{hash}",
     
     ArgumentError:
       unknown keyword: body
     # ./spec/helper/versioning_helper_spec.rb:1305:in `setup_commit_search_stubs'
     # ./spec/helper/versioning_helper_spec.rb:153:in `block (3 levels) in <top (required)>'
     # ./vendor/bundle/ruby/2.6.0/gems/webmock-3.25.1/lib/webmock/rspec.rb:39:in `block (2 levels) in <top (required)>'
```

This didn't happen to me locally because I use Ruby 3.2.0 in my machine.

vegaro (approved):

PR #89, opened 2025-09-30
Insert changelog of older version release in main

ajpallares:
When making a new release of a previous major, we need to insert the changelog of that release into `main`'s changelog. 

This PR adds a new `InsertChangelogOfOlderVersionAction` just for that: if the release corresponds to a previous version than the latest release, a PR is created into `main` that inserts the changelog of the released version into `main`'s changelog at the right position


lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -110,6 +110,29 @@ def self.run(params)
 
           Helper::RevenuecatInternalHelper.create_pr(pr_title, body, repo_name, current_branch, new_branch_name, github_pr_token, [label])
         end
+
+        # For older releases than the latest published version, we need to create a new PR against main to update the changelog with the changes from this release
+        is_older_than_latest_published_version = !Helper::RevenuecatInternalHelper.newer_than_latest_published_version?(new_version_number)
+        if is_older_than_latest_published_version
+          
+          if dry_run
```
vegaro, line 118: should this be if `!dry_run`?

vegaro (approved):
looking good. I think this should be ready to go after tests

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
ajpallares: I _think_ `if dry_run` is correct here:

When `dry_run`, the lines above will generate changes that won't be committed. Therefore, the checkout of `main` will fail unless we discard the previous changes.

When `!dry_run`, all the changes above will have been committed and pushed, so there shouldn't be any uncommited changes at this point.

Does it make sense?

In any case, I will add a comment clarifying this, as it can be confusing 👍

tonidero: Hmm I wonder if we should clear these uncommited changes always on dry-run before this... Since it's not directly related to this process... but yeah, a comment works for now.

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
```diff
@@ -110,6 +110,44 @@ def self.run(params)
...
+        if is_older_than_latest_published_version
+
+          if dry_run
+            Helper::RevenuecatInternalHelper.discard_changes_in_current_branch
+          else
+            UI.important("Creating a new PR against main to update the changelog with the changes from this release")
+          end
+
+          main_branch = "main"
+          Helper::RevenuecatInternalHelper.create_or_checkout_branch(main_branch)
```
tonidero, line 125: Hmm TBH, I'm wondering if we should split this into its own action, so we can call it separately from the rest of the bump action (possibly in a separate job?)... that way we don't need to explicitly checkout main here nor cleanup previous leftovers from the process... Wdyt? 

This could also be moved as part of the release process, instead of the creation of the PR, in case we recreate the release PR multiple times for any reason. So only do this once a release is actually out.

tonidero (commented):
Just some questions about whether this would be better to be performed after an actual release. Looking good otherwise!

lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb
ajpallares: That's a very good point. I did it tis way because it followed the same idea of opening the release PR before actually making the release. But it's true that making it its own separate action that happens right after a release is much cleaner. 
Thanks for the suggestion!

lib/fastlane/plugin/revenuecat_internal/actions/insert_changelog_of_older_version_action.rb
```diff
@@ -0,0 +1,125 @@
...
+            false,
+            hybrid_common_version,
+            sdk_version
+          )
+
+          Helper::RevenuecatInternalHelper.insert_old_version_changelog_in_main(final_version_number, changelog_content, changelog_path)
+
+          if dry_run
+            updated_changelog = File.read(changelog_path)
+            UI.important("The updated changelog would look like this:\n#{updated_changelog}")
```
tonidero, line 49: Hmm won't this print the whole changelog, which can be pretty huge? I'm not sure if it's necessary or whether we should just print what version was added...

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
```diff
@@ -77,22 +84,70 @@ def self.write_changelog(prepopulated_changelog, changelog_latest_path)
...
           version_header = "## #{version_number}"
-          whole_file_data = "#{version_header}\n#{current_changelog_data}\n#{master_changelog_data}"
+          whole_file_data = "#{version_header}\n#{current_changelog_data}\n#{main_changelog_data}"
+
+          main_changelog_write_mode.write(whole_file_data)
+        end
+      end
+
+      def self.insert_old_version_changelog_in_main(version_number, old_version_changelog_contents, changelog_path)
+        main_changelog_data = File.read(changelog_path)
```
tonidero, line 106: This assumes we're in the `main` branch right? Might be worth documenting it on this function... Otherwise, it could make sense to rename to `insert_old_version_changelog_in_current_branch` or something like that?

lib/fastlane/plugin/revenuecat_internal/actions/insert_changelog_of_older_version_action.rb
```diff
@@ -0,0 +1,125 @@
...
+                                       description: "Path to CHANGELOG.latest.md file",
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :repo_name,
+                                       env_name: "RC_INTERNAL_REPO_NAME",
+                                       description: "Name of the repo of the SDK",
+                                       optional: false,
+                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :base_branch,
+                                       description: "The branch you want the CHANGELOG pulled into",
```
tonidero, line 95: Should we default this parameter to `main`?

lib/fastlane/plugin/revenuecat_internal/actions/insert_changelog_of_older_version_action.rb
```diff
@@ -0,0 +1,125 @@
...
+    class InsertChangelogOfOlderVersionAction < Action
+      def self.run(params)
+        sdk_version = params[:sdk_version]
+        changelog_path = params[:changelog_path]
+        changelog_latest_path = params[:changelog_latest_path]
+        repo_name = params[:repo_name]
+        base_branch = params[:base_branch]
+        github_pr_token = params[:github_pr_token]
+        hybrid_common_version = params[:hybrid_common_version]
+        append_phc_version = params[:append_phc_version]
```
tonidero, line 16: To confirm, this will be true for hybrids, null for natives and PHC, correct?

lib/fastlane/plugin/revenuecat_internal/actions/insert_changelog_of_older_version_action.rb
```diff
@@ -0,0 +1,125 @@
...
+
+          current_branch = Actions.git_branch
+
+          changelog_content = File.read(changelog_latest_path)
+
+          if dry_run
+            UI.important("Dry run mode enabled. No changes will be made.")
+          end
+
+          Helper::RevenuecatInternalHelper.create_or_checkout_branch(base_branch)
```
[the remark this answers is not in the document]
ajpallares: Very good point, yes! I can just raise an error if the repo is in a dirty state

lib/fastlane/plugin/revenuecat_internal/actions/insert_changelog_of_older_version_action.rb
tonidero: Right... Maybe it's ok then... We could maybe print the result of `git diff`... but this is probably ok for now 👍

lib/fastlane/plugin/revenuecat_internal/helper/revenuecat_internal_helper.rb
ajpallares: Done in https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/89/commits/45bb28bbb45c564ed4a19570961c397cc0763453

PR #129, opened 2026-04-17
Add safeguards against helper method rename/typo bugs

ajpallares:
## Summary

Adds two complementary safeguards against method rename/typo bugs:

1. **`verify_partial_doubles`** (test-side) — Enabled globally in RSpec. Catches tests that mock a method that doesn't exist on the real class (e.g., a test stubs `find_open_pr_number` but the real method was renamed to `find_unique_open_pr_number`).

2. **Contract test** (implementation-side) — Scans all action source files for calls to any Helper class (`GitHubHelper`, `RevenuecatInternalHelper`, `VersioningHelper`, `UpdateHybridsVersionsFileHelper`, `MavenCentralHelper`) and verifies each called method actually exists. Catches renames/typos in the action code itself that unit tests miss because mocks don't execute the real code path.

### Problem

When `find_open_pr_number` was renamed to `find_unique_open_pr_number`, the unit tests continued passing because RSpec mocks don't verify that stubbed methods exist on the real class by default. This allowed a broken action to ship ([fix commit](https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/commit/46ec2a6910ee4227657534c66f5bdd309bc99688)).

### Changes

- **`spec_helper.rb`**: Enable `verify_partial_doubles = true`. Define `respond_to_missing?` on `FastlaneCore::UI` for its `method_missing`-delegated methods so RSpec can verify them.
- **`spec/contracts/helper_method_contract_spec.rb`**: New test that dynamically scans action files and verifies all helper method calls exist (57 checks across 5 helpers).

### CI verification

Both safeguards were tested on CI by intentionally introducing method rename bugs:

- **Test-side mismatch** ([commit `6317c6b`](https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/commit/6317c6b)) — `verify_partial_doubles` caught it: [Pipeline 590 (failed)](https://app.circleci.com/pipelines/gh/RevenueCat/fastlane-plugin-revenuecat_internal/590)
- **Implementation-side mismatch** ([commit `b1854e4`](https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/commit/b1854e4)) — contract test caught it: [Pipeline 591 (failed)](https://app.circleci.com/pipelines/gh/RevenueCat/fastlane-plugin-revenuecat_internal/591)

Both were reverted in [`54baa28`](https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/commit/54baa28). Full suite passes: 506 examples, 0 failures.


spec/contracts/helper_method_contract_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      helper_methods = helper_class.methods(false).map(&:to_s).to_set
+
+      describe "#{action_basename} -> #{helper_name}" do
+        called_methods.each do |method_name|
+          it "calls .#{method_name} which exists" do
+            expect(helper_methods).to(
+              include(method_name),
+              "#{action_basename} calls #{helper_name}.#{method_name} but that method does not exist. " \
+              "Available public methods: #{helper_methods.sort.join(', ')}"
```
tonidero, line 37: I guess this doesn't verify parameters right? Still I think this might be ok as a first step!!

tonidero (approved):
Thank you for doing this! 🙇 Having mocks for languages like ruby is indeed dangerous 😬

spec/contracts/helper_method_contract_spec.rb
ajpallares: Good point! I missed that.

As you say, that can totally be a next step 👍

PR #130, opened 2026-04-20
Add include_purchases_js option to hybrid changelog generation

tonidero:
## Summary

- Adds an optional `include_purchases_js` parameter (default `false`) to `bump_version_update_changelog_create_pr_action`.
- When enabled, the auto-generated changelog appends links to `purchases-js` releases pulled in by the PHC bump, alongside the existing `purchases-ios` and `purchases-android` links.
- The previous JS version is derived from the previous PHC tag's `purchases-js-hybrid-mappings/package.json`, so no `VERSIONS.md` schema change is needed — only the hybrids that opt in (`purchases-flutter`, `react-native-purchases`) pay for the extra lookup.
- If the previous PHC predates the `purchases-js-hybrid-mappings/` directory or the file can't be parsed, the JS section is skipped gracefully (error logged, iOS/Android links still rendered).

Hybrids that don't support web (cordova, capacitor, unity, kmp) are unaffected — the flag defaults off.

## Test plan

- [x] `bundle exec rspec` — 510 examples, 0 failures
- [x] `bundle exec rubocop` — 73 files, 0 offenses
- [ ] Downstream dry-run in `purchases-flutter` and/or `react-native-purchases` once published (pass `include_purchases_js: true` in the Fastfile and confirm `* [Web x.y.z](...)` appears under the `[AUTOMATIC BUMP] Updates purchases-hybrid-common` entry)

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/update_hybrids_versions_file_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/update_hybrids_versions_file_helper.rb
@@ -1,4 +1,5 @@
 require 'base64'
+require 'json'
 require 'fastlane_core/ui/ui'
 require 'fastlane/action'
 require 'fastlane/actions/github_api'
@@ -24,6 +25,16 @@ module Fastlane
         matches = contents.match("s.dependency 'RevenueCat', '(.*)'").captures
         UI.user_error!("Could not find ios version in #{repo_name} in file '#{path}'") if matches.length != 1
         matches[0]
+      end
+
+      def self.get_js_version_for_hybrid_common_version(hybrid_common_version, github_token)
+        path = 'purchases-js-hybrid-mappings/package.json'
+        repo_name = 'purchases-hybrid-common'
+        contents = get_contents_file_github(path, repo_name, hybrid_common_version, github_token)
+        package_json = JSON.parse(contents)
+        version = package_json.dig('dependencies', '@revenuecat/purchases-js')
+        UI.user_error!("Could not find purchases-js version in #{repo_name} in file '#{path}'") if version.nil? || version.empty?
+        version
       end
 
       def self.get_android_billing_client_version(android_version, github_token)
--- a/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
@@ -487,26 +487,26 @@ module Fastlane
         native_dependency_changelogs += platform_changelogs(ios_releases, 'iOS')
         if include_purchases_js
           previous_phc_version = extract_semver_from_versions_file(versions_latest_release[PHC_VERSION_COLUMN])
-          native_dependency_changelogs += js_releases_links(github_token, phc_version, previous_phc_version)
+          native_dependency_changelogs += js_releases_links(github_token, previous_phc_version, phc_version)
         end
         native_dependency_changelogs.join("\n")
       end
 
-      private_class_method def self.js_releases_links(github_token, phc_version, previous_phc_version)
+      private_class_method def self.js_releases_links(github_token, previous_phc_version, new_phc_version)
         unless Gem::Version.correct?(previous_phc_version)
-          UI.error("Malformed previous PHC version #{previous_phc_version} for version #{phc_version} of purchases-hybrid-common. Skipping purchases-js changelog.")
+          UI.error("Malformed previous PHC version #{previous_phc_version} for version #{new_phc_version} of purchases-hybrid-common. Skipping purchases-js changelog.")
           return []
         end
 
         previous_js_version = Helper::UpdateHybridsVersionsFileHelper.get_js_version_for_hybrid_common_version(previous_phc_version, github_token)
         UI.message("Obtained purchases-js version #{previous_js_version} for previous PHC version #{previous_phc_version}")
-        new_js_version = Helper::UpdateHybridsVersionsFileHelper.get_js_version_for_hybrid_common_version(phc_version, github_token)
-        UI.message("Obtained purchases-js version #{new_js_version} for PHC version #{phc_version}")
+        new_js_version = Helper::UpdateHybridsVersionsFileHelper.get_js_version_for_hybrid_common_version(new_phc_version, github_token)
+        UI.message("Obtained purchases-js version #{new_js_version} for PHC version #{new_phc_version}")
 
         js_releases = Helper::GitHubHelper.get_releases_between_tags(github_token, previous_js_version, new_js_version, REPO_NAME_JS)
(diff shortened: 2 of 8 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -485,9 +485,31 @@ def self.append_phc_version_if_necessary(append_phc_version_if_next_version_is_n
         native_dependency_changelogs = [""]
         native_dependency_changelogs += platform_changelogs(android_releases, 'Android')
         native_dependency_changelogs += platform_changelogs(ios_releases, 'iOS')
+        if include_purchases_js
+          previous_phc_version = extract_semver_from_versions_file(versions_latest_release[PHC_VERSION_COLUMN])
```
tonidero, line 489: Note this is a bit different from how it works in ios/android since we don't have a js column in the VERSIONS.md file right now. I plan to add that in follow-up PRs but this approach should also work and be independent of that.

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -485,9 +485,31 @@ def self.append_phc_version_if_necessary(append_phc_version_if_next_version_is_n
...
+
+        previous_js_version = Helper::UpdateHybridsVersionsFileHelper.get_js_version_for_hybrid_common_version(previous_phc_version, github_token)
+        UI.message("Obtained purchases-js version #{previous_js_version} for previous PHC version #{previous_phc_version}")
+        new_js_version = Helper::UpdateHybridsVersionsFileHelper.get_js_version_for_hybrid_common_version(phc_version, github_token)
+        UI.message("Obtained purchases-js version #{new_js_version} for PHC version #{phc_version}")
+
+        js_releases = Helper::GitHubHelper.get_releases_between_tags(github_token, previous_js_version, new_js_version, REPO_NAME_JS)
+        platform_changelogs(js_releases, 'Web')
+      rescue StandardError => e
+        UI.error("Could not resolve purchases-js versions for PHC #{previous_phc_version} -> #{phc_version}: #{e.message}. Skipping purchases-js changelog.")
```
ajpallares: Hmm I'm seeing that this could cover up some errors of our own.

For example, in `get_js_version_for_hybrid_common_version` we raise a `UI.user_error!` 
```
UI.user_error!("Could not find purchases-js version in #{repo_name} in file '#{path}'") if version.nil? || version.empty?
```

but that error would be swallowed by this if I'm not reading wrong.

Not a blocker, since in any case this is an improvement with respect to what we have right now, but I wonder if we should be more strict here.

lib/fastlane/plugin/revenuecat_internal/helper/update_hybrids_versions_file_helper.rb
```diff
@@ -26,6 +27,16 @@ def self.get_ios_version_for_hybrid_common_version(hybrid_common_version, github
         matches[0]
       end
 
+      def self.get_js_version_for_hybrid_common_version(hybrid_common_version, github_token)
```
ajpallares, line 30: May be worth adding some tests for this method

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
```diff
@@ -485,9 +485,31 @@ def self.append_phc_version_if_necessary(append_phc_version_if_next_version_is_n
...
         native_dependency_changelogs += platform_changelogs(android_releases, 'Android')
         native_dependency_changelogs += platform_changelogs(ios_releases, 'iOS')
+        if include_purchases_js
+          previous_phc_version = extract_semver_from_versions_file(versions_latest_release[PHC_VERSION_COLUMN])
+          native_dependency_changelogs += js_releases_links(github_token, phc_version, previous_phc_version)
+        end
         native_dependency_changelogs.join("\n")
       end
 
+      private_class_method def self.js_releases_links(github_token, phc_version, previous_phc_version)
```
ajpallares: AI Nit: could we swap the argument order to `js_releases_links(github_token, previous_phc_version, new_phc_version)`? The rest of the file consistently uses "previous → new" (e.g. `get_releases_between_tags(..., previous_ios_version, new_ios_version, ...)`), so this signature is the odd one out and a bit tripping to read.

ajpallares (approved):
Makes sense to me! Some small, non-blocking comments. Thank you for adding this!

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
tonidero: Great point! I considered this step optional... But TBH, If it fails here, it might be worth failing the command I think. Will bump it to user_error. Thank you!!

ajpallares: Right. I see why you made it optional. But yeah, let's see how this works out and if it gets too noisy, we can iterate 👍 
Thanks!

lib/fastlane/plugin/revenuecat_internal/helper/update_hybrids_versions_file_helper.rb
tonidero: Added in https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/130/commits/a256c3666f11ab4b8bc550b7870d2c4df774b641

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
tonidero: Fixed in https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/130/commits/ff77793dd022ad64fadec5367a2642367d32930d

lib/fastlane/plugin/revenuecat_internal/helper/versioning_helper.rb
tonidero: Fixed in https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/130/commits/b84dd9d16ff051e159e1b645fcdaa5b34cb756d6

spec/test_files/get_contents_phc_js_hybrid_mappings.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "package.json",
+    "path": "purchases-js-hybrid-mappings/package.json",
+    "sha": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```
ajpallares, line 4: 😄

ajpallares (approved):
Thanks for the new changes! 🙌

PR #133, opened 2026-04-30
Disambiguate multi-PR search results by merge_commit_sha

ajpallares:
## Why

GitHub's `SHA:<sha>` PR-search qualifier matches **any PR whose branch contains the commit**, not just the PR that introduced it. When a stacked PR brings the base branch into its head, every commit pulled in matches that stacked PR plus its real originating PR — and `get_pr_resp_items_for_sha` returns both, so `auto_generate_changelog` and `determine_next_version_using_labels` bail out with `Multiple commits found for <sha>`. Empty `base_branch` (e.g. detached-HEAD CI on a tag) makes this strictly more likely.

Reproduced on `purchases-ios` `5.70.0` (failure [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/37252/workflows/a2fe4367-7a8d-41ab-a17c-de7dc83e5d04/jobs/552869), originating PR [#6693](https://github.com/RevenueCat/purchases-ios/pull/6693), spurious match [#6697](https://github.com/RevenueCat/purchases-ios/pull/6697)).

## What

When the search returns >1 PRs for the same SHA, fetch each candidate's `merge_commit_sha` (the commit GitHub leaves on the base ref — works for squash, merge-commit, and rebase merges) and keep only PRs whose merge SHA equals the queried commit. If exactly one survives, return it; otherwise return the original list so the caller errors out with an actionable message that lists the ambiguous PR numbers.

Also logs a warning when `base_branch` is empty.

Single-result happy path: zero extra API calls. Public method signatures and behavior on previously-working inputs are unchanged.

## Tests

`bundle exec rake` (rspec + rubocop): **546 examples, 0 failures, no offenses.** New positive specs in `github_helper_spec.rb` for each disambiguation branch; the existing multi-result specs in `versioning_helper_spec.rb` were updated to stub the disambiguation request and assert the new error message.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
@@ -40,6 +40,13 @@ module Fastlane
       # When it is a non-nil string, and the search API returns no items, the helper
       # attempts a fallback lookup by extracting the PR number from the commit message
       # and fetching the PR directly via the REST API. Pass nil to disable the fallback.
+      #
+      # When the search API returns more than one PR for the same SHA (e.g. because a
+      # stacked PR brought commits from the base branch into its head), this helper
+      # disambiguates by fetching each candidate's `merge_commit_sha` and keeping only
+      # the PR whose merge SHA equals the queried commit. If exactly one candidate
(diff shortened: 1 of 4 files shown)


tonidero (approved):
Thanks for fixing this!!

PR #134, opened 2026-05-01
fix: retry pod push on stale CocoaPods spec index

facumenzella:
## Summary

- Adds `"None of your spec sources contain a spec satisfying the dependency"` as a retryable error in `pod_push_with_error_handling`. This handles the race condition where a dependency podspec (e.g. `PurchasesHybridCommon`) was just pushed but the CocoaPods index hasn't propagated to CI yet — causing `PurchasesHybridCommonUI` push to fail validation.
- Logs an explicit message for unknown (non-retryable) errors: `"❌ Pod push failed with an unknown error and won't retry. You can rerun this job using SSH."`
- Adds test coverage for the stale spec index retry case.

## Test plan

- [ ] All existing specs pass (`bundle exec rspec spec/actions/pod_push_with_error_handling_action_spec.rb`)
- [ ] New test covers the stale spec index retry scenario

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/pod_push_with_error_handling.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/pod_push_with_error_handling.rb
@@ -46,6 +46,7 @@ module Fastlane
               return false
             end
 
+            FastlaneCore::UI.error("❌ Pod push failed with an unknown error and won't retry. You can rerun this job using SSH. Error: #{e.message}")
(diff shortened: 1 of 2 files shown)


rickvdl (approved):
Nice, good idea. Question: do we have any idea if this delay is enough? IIRC sometimes it could take minutes even I think...

facumenzella:
@rickvdl not really :( 
I think we've started like this and it kind of worked. I don't even remember when we added this 🤔

PR #137, opened 2026-06-11
Stop (double) requesting reviewers for PHC bump PRs

rickvdl:
As our team is already listed in the codeowners files we were getting duplicate review requests. NABD but this should no longer request our team's review specifically as well.


tonidero (approved):
Right, makes sense! Thank you!

ajpallares (approved):

PR #139, opened 2026-06-29
Render on-call Slack mention via Block Kit so it works on mobile

ajpallares:
### Motivation

The integration-test feed alert pings the SDK on-call group, but on the Slack iOS app the mention shows as `@[group unavailable]` (it only resolves on desktop). Root cause: fastlane's `slack` action posts the message inside a [legacy secondary attachment](https://docs.slack.dev/legacy/legacy-messaging/legacy-secondary-message-attachments), where Slack renders mentions unreliably and content "may be wrapped, truncated, or hidden" by clients. Slack [recommends migrating to Block Kit](https://docs.slack.dev/messaging/migrating-outmoded-message-compositions-to-blocks/).

### Summary

- Post a Block Kit payload to the webhook directly instead of via `other_action.slack`.
- Headline + on-call mention now live in a top-level block (renders reliably on mobile); run details stay in a Block-Kit-based colored attachment.
- No public API change — same action parameters and options.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/actions/slack_backend_integration_test_results.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/actions/slack_backend_integration_test_results.rb
@@ -41,15 +41,15 @@ module Fastlane
 
         post_messages(
           success: success,
-          notify_binary_solo: message_binary_solo_on_failure,
+          notify_binary_solo_on_failure: message_binary_solo_on_failure,
           feed_message: message_feed,
           fields: detail_fields,
           build_url: build_url
         )
       end
 
-      def self.post_messages(success:, notify_binary_solo:, feed_message:, fields:, build_url:)
-        if notify_binary_solo && !success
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/revenuecat_internal/actions/slack_backend_integration_test_results.rb
```diff
@@ -1,42 +1,25 @@
...
+require 'uri'
+require 'json'
 require_relative '../helper/revenuecat_internal_helper'
 
 module Fastlane
   module Actions
     class SlackBackendIntegrationTestResultsAction < Action
       ON_CALL_SDK_MENTION = "<!subteam^S0939BTV0SY|oncall-sdk>"
 
-      # rubocop:disable Metrics/PerceivedComplexity
```
ajpallares, line 10: Took the opportunity to remove this line by extracting some methods

tonidero (approved):
:shipit: Thank you!!

PR #143, opened 2026-07-16
Check CocoaPods trunk when a pod push errors

ajpallares:
### What & why
`pod_push_with_error_handling` detected an already-published pod by matching `[!] Unable to accept duplicate entry for:` against the raised error's message. Since fastlane 2.237.0, `sh` truncates command output and omits it from `FastlaneShellError` messages when stdout is a TTY, so that string is no longer present — an already-published pod (e.g. re-running a release after a partial failure) now fails the job.

This queries CocoaPods trunk directly instead: on a push error, if the podspec's version is already published, the push is treated as successful. Successful pushes skip the check. The transient-error retry path is unchanged.

Note: this fixes the "already published" case only. Reliable retry-classification for transient errors (also affected by the same 2.237.0 change) is a separate follow-up.

- [x] Unit tests


facumenzella (approved):
nice one

PR #145, opened 2026-07-17
fix: resolve android billing client version across toml key names

ajpallares:
### Motivation

Hybrid version bumps crashed in `update_hybrids_versions_file` with `undefined method 'captures' for nil:NilClass` (seen in [purchases-capacitor bump](https://app.circleci.com/pipelines/github/RevenueCat/purchases-capacitor/2363/workflows/30f048a9-ff17-4dfd-b9f1-3609daf169ec/jobs/16945)). `get_android_billing_client_version` hardcoded a `bc8 = "..."` lookup in purchases-android's `gradle/libs.versions.toml`, but that key was removed when android consolidated `bc7`/`bc8` back to `billingClient` (purchases-android#3717). The failed match returned `nil`, and `.captures` raised instead of the intended error.

### Summary

- Match `billingClient` first, fall back to `bc8`, so the version resolves regardless of which key name a given android release uses.
- Guard the `nil` match so a genuinely missing key raises the intended `user_error!` instead of an opaque `NoMethodError`.


tonidero (approved):
Ouch forgot about this one. Thank you!!

PR #148, opened 2026-07-31
Attribute commits from stacked PRs when generating changelogs

ajpallares:
### Motivation
The PR search is scoped to `base:<branch>`, so it never matches a PR that targeted another branch — every PR in a stack except the bottom one, and PRs merged into a long-lived dev branch. Merging a stack lands one squashed commit per PR on the base branch, and each fell through to the "no PR found" path: no author, no labels, and (in repos that squash with `PR_BODY`) the whole PR description copied into the changelog. Losing the labels also meant those PRs were ignored when determining the next version, so a stacked `pr:feat` would ship as a patch. Reported in [#sdk-team](https://revenuecat.slack.com/archives/C03HPN9UE80/p1785507590109059) after a stack landed in purchases-unity and the changelog had to be [fixed by hand](https://github.com/RevenueCat/purchases-unity/pull/1034/commits/0a280fab0a533bdd042d2f002440b93d1716f392).

### Summary
- When the search finds nothing, look up the commit's associated PRs and keep only merged PRs whose `merge_commit_sha` is the commit itself.

### Notes
- That filter is what keeps a PR merged into someone's feature branch out of the changelog: its merge commit is squashed away and never reaches main. Verified against five such PRs across three repos.
- Applies with no opt-in, unlike `fallback_pr_lookup`, which stays for merge-queue commits.
- Verified against the four real purchases-unity stack commits, which now resolve to #836–#897 with their labels.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/github_helper.rb
@@ -36,10 +36,14 @@ module Fastlane
         end
       end
 
-      # When +fallback_commit_message+ is nil, the search API result is returned as-is.
-      # When it is a non-nil string, and the search API returns no items, the helper
-      # attempts a fallback lookup by extracting the PR number from the commit message
-      # and fetching the PR directly via the REST API. Pass nil to disable the fallback.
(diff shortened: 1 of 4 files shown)


tonidero (approved):
🚢 Thank you!!

PR #149, opened 2026-08-03
Add `upload_snapshots_to_sentry` action

joshdholtz:
## Problem

iOS and Android each carry their own copy of the Sentry Snapshots upload logic in their Fastfiles, and the copies have already drifted once: a diff parsing bug got fixed on the Android side while iOS still has it.

## Fix

One `upload_snapshots_to_sentry` action that owns everything after image generation. It refuses partial runs that would wreck the main baseline, uploads only the images that changed (Sentry bills per image), and falls back to a full upload when there is no baseline yet. The app repos keep their own generation (Emerge on iOS, Paparazzi on Android) and their lanes shrink to generate plus one action call.

## How was this tested?

New specs cover the guards, the selective staging, and the fallback paths against captured output from a real sentry-cli diff. Full suite passes and rubocop is clean.

### DIFF
--- a/lib/fastlane/plugin/revenuecat_internal/helper/sentry_snapshots_helper.rb
+++ b/lib/fastlane/plugin/revenuecat_internal/helper/sentry_snapshots_helper.rb
@@ -1,0 +1,121 @@
+require 'fastlane_core/ui/ui'
+require 'fastlane/action'
+require 'fileutils'
+require 'json'
+require 'tmpdir'
+
+module Fastlane
+  UI = FastlaneCore::UI unless Fastlane.const_defined?(:UI)
+
+  module Helper
+    # Uploads a directory of snapshot images to Sentry Snapshots: refuses
+    # partial runs that would corrupt the main baseline, uploads only images
+    # that changed vs that baseline (Sentry bills per image), and falls back
+    # to a full upload when no baseline exists or the diff fails.
+    class SentrySnapshotsHelper
+      CHANGED_STATUSES = %w[changed added].freeze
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/revenuecat_internal/helper/sentry_snapshots_helper.rb
```diff
@@ -0,0 +1,129 @@
...
+        UI.user_error!("Generated #{generated} snapshots; the #{main_branch} baseline has #{baseline}. " \
+                       "Refusing to upload what looks like a partial run. Set SNAPSHOT_COUNT_OVERRIDE=1 for intentional removals.")
+      end
+
+      # Stages only changed/added images (with their .json sidecars, which
+      # carry display names and tags) for a --selective upload. Returns
+      # [upload_dir, all_names_file], or nil on any failure (callers then
+      # full-upload).
+      def self.select_changed_snapshots(export_dir, app_id, sentry_cli, main_branch)
+        work_dir = Dir.mktmpdir("snapshots")
```
fire-at-will, line 67: I don't think that this snapshots directory is ever cleaned up. That probably isn't a big deal when we're running on temporary CI machines, but might be worth cleaning it up for completeness and in case someone ever runs this on their own machine

fire-at-will (approved):

lib/fastlane/plugin/revenuecat_internal/helper/sentry_snapshots_helper.rb
joshdholtz: Correct in that its never cleaned up in this job! It does get put into the OS temp dir so it will be gone eventually 🙃

